PR #4, opened 2015-06-05
Add URI factory (fix #3)

### LINKED ISSUE #3: Guesser for UriFactory
And probably UriFactory implementations for Guzzle and Diactoros?

ddeboer:
Fix #3.

Also refactored a bit to reduce code duplication.

### DIFF
--- a/src/ClassDiscovery.php
+++ b/src/ClassDiscovery.php
@@ -1,0 +1,51 @@
+<?php
+
+namespace Http\Discovery;
+
+/**
+ * Registry that based find results on class existence
+ *
+ * @author David de Boer <david@ddeboer.nl>
+ */
+abstract class ClassDiscovery
+{
+    /**
+     * Add a condition (and class) to the discovery registry
+     *
+     * @param string $name
+     * @param string $class     Class that will be instantiated if found
+     * @param string $condition Optional other class to check for existence
+     */
+    public static function register($name, $class, $condition = null)
+    {
+        static::$cache = null;
+
+        static::$classes[$name] = [
+            'class' => $class,
+            'condition' => $condition ?: $class,
+        ];
+    }
+
+    /**
+     * Finds a Factory
+     *
+     * @return object
+     *
+     * @throws NotFoundException
+     */
+    public static function find()
+    {
+        // We have a cache
+        if (isset(static::$cache)) {
+            return static::$cache;
+        }
+
+        foreach (static::$classes as $name => $definition) {
+            if (class_exists($definition['condition'])) {
+                return static::$cache = new $definition['class'];
+            }
+        }
+
+        throw new NotFoundException('Not found');
+    }
+}
--- a/src/HttpAdapterDiscovery.php
+++ b/src/HttpAdapterDiscovery.php
@@ -11,59 +11,47 @@
 
 namespace Http\Discovery;
 
+use Http\Adapter\HttpAdapter;
+
 /**
  * Finds an HTTP Adapter
  *
  * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
  */
-class HttpAdapterDiscovery
+class HttpAdapterDiscovery extends ClassDiscovery
 {
     /**
-     * @var array
-     */
-    protected static $adapters = [
-        'guzzle6' => 'Http\Adapter\Guzzle6HttpAdapter',
-        'guzzle5' => 'Http\Adapter\Guzzle5HttpAdapter',
-    ];
-
-    /**
-     * @var string
+     * Cached adapter
+     *
+     * @var HttpAdapter
      */
     protected static $cache;
 
     /**
-     * Register an HTTP Adapter
-     *
-     * @param string $name
-     * @param string $class
+     * @var array
      */
-    public static function register($name, $class)
-    {
-        static::$cache = null;
+    protected static $classes = [
+        'guzzle6' => [
+            'class' => 'Http\Adapter\Guzzle6HttpAdapter',
+            'condition' => 'Http\Adapter\Guzzle6HttpAdapter'
 
-        static::$adapters[$name] = $class;
-    }
+        ],
+        'guzzle5' => [
+            'class' => 'Http\Adapter\Guzzle5HttpAdapter',
+            'condition' => 'Http\Adapter\Guzzle6HttpAdapter'
+        ],
+    ];
 
     /**
      * Finds an HTTP Adapter
      *
-     * @return object
+     * @return HttpAdapter
      *
      * @throws NotFoundException
      */
     public static function find()
     {
-        // We have a cache
-        if (isset(static::$cache)) {
-            return static::$cache;
-        }
-
-        foreach (static::$adapters as $name => $class) {
-            if (class_exists($class)) {
-                return static::$cache = new $class;
-            }
-        }
-
-        throw new NotFoundException('No HTTP Adapter found');
+        // Override only used for return type declaration
+        return parent::find();
     }
 }
--- a/src/UriFactory/GuzzleFactory.php
+++ b/src/UriFactory/GuzzleFactory.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Http\Discovery\UriFactory;
+
+use GuzzleHttp\Psr7;
+use Http\Message\UriFactory;
+
+/**
+ * Creates a guzzlehttp/psr7 URI object
+ *
+ * @author David de Boer <david@ddeboer.nl>
+ */
+class GuzzleFactory implements UriFactory
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function createUri($uri)
+    {
+        return Psr7\uri_for($uri);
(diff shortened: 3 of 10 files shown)


src/ClassDiscovery.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+namespace Http\Discovery;
+
+/**
+ * Registry that based find results on class existence
+ *
+ * @author David de Boer <david@ddeboer.nl>
+ */
+abstract class ClassDiscovery implements Discovery
```
sagikazarmark: I would rather call it FactoryDiscover, it tells more about it's purpose IMO.

src/ClassDiscovery.php
```diff
@@ -0,0 +1,58 @@
...
+{
+    protected static $cache;
```
sagikazarmark: Some docblocks are missing.

src/ClassDiscovery.php
ddeboer: True, but I thought we could make HttpAdapter extend the abstract class too.

sagikazarmark: This isn't too much code that complexity worths avoiding duplication. The guess condition logic is a bit different there.

ddeboer: I already took that into account in the abstract class.

sagikazarmark: Fine, but in that case the parameters should always be the same. See comment bellow.

src/ClassDiscovery.php
```diff
@@ -0,0 +1,58 @@
...
+    public static function find()
+    {
+        // We have a cache
+        if (isset(static::$cache)) {
+            return static::$cache;
+        }
+
+        foreach (static::$factories as $name => $definition) {
+            if (class_exists($definition['class'])) {
+                if (isset($definition['factory'])) {
```
sagikazarmark: I would rather call call class the one which should be instantiated and condition which should be checked. This way the target class is consistent between all guessers.

ddeboer: Good idea.

src/Discovery.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Http\Discovery;
+
+interface Discovery
```
sagikazarmark: I wonder if this interface is really necessary. Since we are in static context, you can't even check the instance. So all Discovery objects are treated as final static classes. WDYT?

src/ClassDiscovery.php
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Registry that based find results on class existence
+ *
+ * @author David de Boer <david@ddeboer.nl>
+ */
+abstract class ClassDiscovery implements Discovery
+{
+    protected static $cache;
+    protected static $factories = [];
```
sagikazarmark: This could be called targets or something less factory specific.

src/Discovery.php
ddeboer: Yep, dropping the interface.

sagikazarmark: Anyway, the idea of communicating that they are something similar is good. But interface cannot be really used here.

src/HttpAdapterDiscovery.php
```diff
@@ -11,19 +11,28 @@
...
     /**
      * @var array
      */
-    protected static $adapters = [
-        'guzzle6' => 'Http\Adapter\Guzzle6HttpAdapter',
-        'guzzle5' => 'Http\Adapter\Guzzle5HttpAdapter',
+    protected static $classes = [
+        'guzzle6' => [
+            'condition' => 'Http\Adapter\Guzzle6HttpAdapter',
+            'class' => 'Http\Adapter\Guzzle6HttpAdapter'
```
ddeboer: This is duplicated, but I like that it is explicit about that it’s going to return the same class as it finds.

ddeboer:
Updated.

src/ClassDiscovery.php
```diff
@@ -0,0 +1,65 @@
...
+    protected static $cache;
+
+    /**
+     * Conditions and classes
+     *
+     * @var array
+     */
+    protected static $classes = [];
+
+    /**
```
sagikazarmark: They are not factories anymore, are they? 😃

src/ClassDiscovery.php
```diff
@@ -0,0 +1,65 @@
...
+     * Register a factory
+     *
+     * @param string $name
+     * @param string $condition
+     * @param string $class
+     */
+    public static function register($name, $condition, $class = null)
```
sagikazarmark: I would change the order of arguments: the first is the class, which is always required. The second is the condition. Even in factory guessers, it is acceptable to pass the condition as a second argument. Again, just for the sake of consistency.

src/ClassDiscovery.php
ddeboer: Fixed.

src/HttpAdapterDiscovery.php
sagikazarmark: No objections.

src/ClassDiscovery.php
ddeboer: So `$class, $condition = null`?

sagikazarmark: Yup, that seems more logical to me.

ddeboer: By the way, `$name` doesn’t really add much. A priority would be better. But that’s for another PR. :wink:

src/ClassDiscovery.php
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @author David de Boer <david@ddeboer.nl>
+ */
+abstract class ClassDiscovery
+{
+    /**
+     * Cached class
+     *
+     * @var object
+     */
```
sagikazarmark: Be aware, that in this case tha cache property will be the same for all child classes. You either have to add a cache property to all discovery objects or handle this case in the parent, like saving the cache in an array based on get_called_class.

src/ClassDiscovery.php
sagikazarmark: Not sure if priority matters either. It is very unlikely you have two adapters/message implementations installed.

ddeboer: True.

sagikazarmark: To answer your question: two possible way I can see: make condition required argument or check it in cases when required.

For the long term, i would convert the condition to something that can be evaluated as boolean. So we could accept a callable or boolean as well. In that case, null is obviously false.

src/ClassDiscovery.php
sagikazarmark: To be clear: when a static property is declared in the parent, but not the child then setting it in the child will set it in the parent, since there is no context.

ddeboer: Yeah, I overlooked that. Let’s add it to the childs.

src/HttpAdapterDiscovery.php
```diff
@@ -32,38 +41,14 @@ class HttpAdapterDiscovery
...
-        }
-
-        foreach (static::$adapters as $name => $class) {
-            if (class_exists($class)) {
-                return static::$cache = new $class;
-            }
-        }
-
-        throw new NotFoundException('No HTTP Adapter found');
+        return parent::find();
```
sagikazarmark: Are these methods necessary? For the sake of return type?

ddeboer: Necessary, no. But (as you may have noticed elsewhere) I value IDE auto-completion very much.

sagikazarmark: Fine. 😃 But in that case, add a comment so that I dont remove it as auto optimalization. 😃

src/UriFactory/GuzzleFactory.php
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @author David de Boer <david@ddeboer.nl>
+ */
+class GuzzleFactory implements UriFactory
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function createUri($uri)
+    {
```
sagikazarmark: IIRC there is a uri_for function in guzzle. Maybe it worth to use that?

ddeboer: So as to support passing `UriInterface` objects as `$uri`?

sagikazarmark: Yup. The reason why UriInterface is accepted in the factory interface is that you can accept both string and uri implementation in your library and you can leave the check to the factory.

ddeboer: Done. By the way, if you comment on the line itself, the discussion will be cleaned up if that line changes.)

sagikazarmark: Yeah, my bad

ddeboer:
Okay, should be ready now.

sagikazarmark:
Thanks

ddeboer:
You too, for the thorough review. :smile:

PR #18, opened 2015-11-10
Add StreamFactoryDiscovery

mekras:
Supports Guzzle and Zend Diactoros.

### DIFF
--- a/src/StreamFactory/DiactorosStreamFactory.php
+++ b/src/StreamFactory/DiactorosStreamFactory.php
@@ -20,7 +20,6 @@ use Zend\Diactoros\Stream;
 /**
  * Zend Diactoros StreamFactory
  *
- * @author GeLo <geloen.eric@gmail.com>
  * @author Михаил Красильников <m.krasilnikov@yandex.ru>
  */
 class DiactorosStreamFactory implements StreamFactory
--- a/src/StreamFactory/DiactorosStreamFactory.php
+++ b/src/StreamFactory/DiactorosStreamFactory.php
@@ -18,8 +18,6 @@ use Zend\Diactoros\Response;
 use Zend\Diactoros\Stream;
 
 /**
- * Zend Diactoros StreamFactory
- *
  * @author Михаил Красильников <m.krasilnikov@yandex.ru>
  */
 class DiactorosStreamFactory implements StreamFactory
--- a/src/StreamFactory/DiactorosStreamFactory.php
+++ b/src/StreamFactory/DiactorosStreamFactory.php
@@ -1,0 +1,58 @@
+<?php
+
+/*
+ * This file is part of the Http Discovery package.
+ *
+ * (c) PHP HTTP Team <team@php-http.org>
+ *
+ * For the full copyright and license information, please read the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Http\Discovery\StreamFactory;
+
+use Http\Message\StreamFactory;
+use Psr\Http\Message\StreamInterface;
+use Zend\Diactoros\Request;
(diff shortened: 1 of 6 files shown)


src/StreamFactory/DiactorosStreamFactory.php
```diff
@@ -0,0 +1,58 @@
...
+use Http\Message\StreamFactory;
+use Psr\Http\Message\StreamInterface;
+use Zend\Diactoros\Request;
+use Zend\Diactoros\Response;
+use Zend\Diactoros\Stream;
+
+/**
+ * Zend Diactoros StreamFactory
+ *
+ * @author GeLo <geloen.eric@gmail.com>
```
sagikazarmark: Remove Eric please.

src/StreamFactory/DiactorosStreamFactory.php
```diff
@@ -0,0 +1,58 @@
...
+namespace Http\Discovery\StreamFactory;
+
+use Http\Message\StreamFactory;
+use Psr\Http\Message\StreamInterface;
+use Zend\Diactoros\Request;
+use Zend\Diactoros\Response;
+use Zend\Diactoros\Stream;
+
+/**
+ * Zend Diactoros StreamFactory
```
sagikazarmark: This is not necessary

src/StreamFactory/GuzzleStreamFactory.php
```diff
@@ -0,0 +1,37 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Http\Discovery\StreamFactory;
+
+use Http\Message\StreamFactory;
+use Psr\Http\Message\StreamInterface;
+
+/**
+ * Guzzle Stream Factory
```
sagikazarmark: This is not necessary

sagikazarmark:
Nice job @mekras

src/StreamFactory/DiactorosStreamFactory.php
mekras: But it's code written by him. I just copied it from MessageFactory.

sagikazarmark: Actually, it is written by me. Back then I added Eric to each files he had any connection with, but that class is 90% rewritten by me. Even if it is copied partly, please add only yourself.

You might also add a comment that the code is copied from message factory. Actually, the ultimate solution would be somehow share the code between the two classes. But let's think about it later.

src/StreamFactory/DiactorosStreamFactory.php
mekras: Not necessarily, but also not forbidden, right?

src/StreamFactory/GuzzleStreamFactory.php
mekras: Not necessarily, but also not forbidden, right?

sagikazarmark: No, but I consider it to be noise as there is no added value in it. (It is exactly the same as the classname) I am going to remove it from other factories as well.

src/StreamFactory/DiactorosStreamFactory.php
mekras: OK

joelwurtz:
I think message factory should have an injected stream factory to create stream (actually it create stream from a protected method)

sagikazarmark:
I also thought about that. But it would complicate using the MessageFactory.

Actually, we could even instantiate the stream factory or use the StreamDiscovery in MessageFactory.

Since the MessageFactory uses the exact same code, hard coupling is probably not a problem here.

PR #30, opened 2015-12-16
Extract message factory

sagikazarmark:


dbu:
the lowest version build looks weird. can the guzzle adapter really not support php 5.4? should we just move to 5.5? what feature do we need?

sagikazarmark:
Yeah, I realized. Guzzle 6 only supports PHP 5.5 unfortunately.

I think it would be okay to move the tests to 5.5. This was the reason for the ugly interface hack. Optionally we could require guzzle 5 instead, but as guzzle 6 is our main implementation, that's not so good IMO.

dbu:
well, testing lowest version with guzzle 5 would be interesting to see
if other versions work as well. otherwise we only know about guzzle 6.

ah, but the guzzle 5 adapter is not even done, so maybe lets just leave it.

Issue #28, opened 2015-12-16
Use Puli discovery


sagikazarmark:
I have been looking at puli for the last few weeks. I think @joelwurtz also spent some time with it, at least during PHP Forum with @webmozart.

I think it is quite awesome and we could easily use it not just in the discovery package, but in the whole project. Since we have nice contracts, we already have out binding types: HttpClient, HttpAsyncClient, *Factory. In each of our packages we can provide a puli.json with the actual bindings (like Guzzle6HttpAdapter is bind to HttpClient and HttpAsyncClient)

Downside of this solution is that every resource must be registered in the repository (?) and conditional auto-discovery (and caching) during runtime won't work. However, this is not necessarily a downside, but probably makes flexibility a bit harder.

While this could make our static discoveries obsolete, I would still keep them. Although Puli does not recommend it to use it in global context or instantiate it more than once, I think we should use them inside our discovery layer. The workflow would be the following:

If someone wants to resolve something from the discovery, and it is the first call to any discoveries, we instantiate the Puli factory and store it in the ClassDiscovery object. After that, inside the specific discovery (like HttpClientDiscovery) we loop through our internal list of clients and append them to Puli ONCE. We set an initialized bit or something. This replaces our internal cache mechanism. After that, we resolve the first match of the specific binding and return it.

Possible pitfalls:

- Performance: Puli recommends NOT to this global context thing for performance (and other) reasons. However our discovery thing's aim is zero config, which itself is not recommended. Configuration, DI should be used, but if you need it as is, you can use it. Not sure if we need to care about this apart from placing a big fat warning somewhere.
- Priority: not sure about Puli's priority handling. FIFO? LIFO? Does it really matter? IMO when we use a discovery, we just want something fast, should we really care about priorities? Currently it is possible to add resources to discoveries and they are returned FIFO.
- Puli required: with this we would add puli as a dependency. Not sure if it is really a downside, as puli is quite awesome. Putting a puli.json file in all of the related repos should really not be a problem.

Reference:
https://speakerdeck.com/webmozart/puli-phps-next-package-revolution?slide=102

dbu:
oh, i was not aware that puli can provide objects - i thought its about
resource files and such things. but then, why not ditch discovery
completely and document how to use puli for autodiscovery? the only
thing we would need to provide is a puli.json?

the current discoveries are not really complicated, and they are a last
resort thing, i would like to avoid overcomplicating them. we might also
want to look at how laravel is doing things - the discovery looks a lot
[...]

PR #34, opened 2015-12-25
Use Puli discovery

sagikazarmark:
As discussed #28, this PR aims to replace any custom discovery logic with [Puli](http://puli.io).

The convenient static wrappers are preserved, there is no major API change.

I am pretty in favour of this PR, really like how Puli works.

One arguable thing: this PR requires Puli to be installed on the user's machine or to be installed through composer (as in the require-dev of this package). I am fine with that requirement, even if it is beta, but I would like to hear other opinions about this.

Many thanks to @webmozart for helping me with Puli.


jeromegamez:
Big Puli :+1:!

.gitignore
```diff
@@ -1,5 +1,7 @@
+.puli/
 build/
 vendor/
 composer.lock
 phpspec.yml
 phpunit.xml
+puli.json
```
webmozart: Did you not commit the puli.json file by intention? Just like composer.json, this file should usually be committed and distributed with the package.

sagikazarmark: I generally agree, but this package is supposed to be installed in other packages or an application. There are no bindings or binding types in this package, so only the "lock" functionality would be generated into the json file, which IMO doesn't make sense here, but would make sense in the application.

Why is the package list generated in the json file in the first place? I though the json file is some kind of configuration and I expect it doesn't contain anything else. The generated package list could be part of some kind of lock file, couldn't it?

webmozart: Ok, makes sense then.

The package list is in there since, at least in principle, Puli is independent of Composer. You could install additional packages manually or generally use Puli without Composer, if that makes sense for your project. So puli.json duplicates some parts of composer.json, however those parts are synchronized by the Composer plugin.

sagikazarmark: Actually I was thinking about a puli.lock file. In cases like this, a puli.json might not even be necessary in a project/application, only a puli.lock where non-configuration stuff can go. Even if you add stuff manually there, it is rather an "environmental" (aka. project dependent) thing, which depends on external resources, while you might change a configuration based on internal project changes.

webmozart: Puli works differently than Composer. There's no need for a lock file in Puli.

You are right that in this case, a puli.json is not needed (but neither is a lock file).

sagikazarmark: Well, calling it lock is probably not straightforward and multiple configuration can be confusing. Also, packages for composer are only generated when the composer plugin is installed and it should only be done in an application, am right? With other words: package configuration, like composer packages should only be in an application, but not in reusable packages. Am I correct?

webmozart: Not necessarily. If your packages contain integration tests that require f.i. Puli's Discovery service and certain binding types/bindings to be loaded from other packages, then those packages must be loaded into Puli as well, otherwise the binding types/bindings are not found.

sagikazarmark: I see. Are those packages loaded into the application where the package is installed afterwards? Because I probably don't want integration testing binding in the application where I use the package.

webmozart: No. :)

sagikazarmark: Cool. Thanks for clarification.

ddeboer:
This is a great improvement! Good to use something for discovery that’s already good at discovery, like Puli. :wink:

@sagikazarmark Let’s squash and merge?

PR #37, opened 2016-01-03
make puli requirement more flexible

dbu:


composer.json
```diff
@@ -11,9 +11,9 @@
         }
     ],
     "require": {
-        "php": ">=5.4",
-        "puli/composer-plugin": "1.0.0-beta8",
-        "puli/discovery": "1.0.0-beta8"
+        "php": "^5.4|7.*",
```
sagikazarmark: Is this better than ^7.0?

dbu: its semantically the same. maybe caret is better for when we need to bump
to some php 7 that includes a fix (if we run into that situation). but we
can also change once it becomes relevant

sagikazarmark: I don't mind if it's the same.

sagikazarmark:
:+1:

PR #39, opened 2016-01-04
require puli cli to work out of the box

dbu:
as per the discussion in https://github.com/FriendsOfSymfony/FOSHttpCache/pull/257/files#r48696611

for discovery to work, we need puli. discovery is all about ease of use, mostly a first impression thing (to properly configure your client, you want to instantiate/configure the client yourself). the first impression should be as smooth as possible. puli can't be expected to be globally installed in most systems. the worst that can happen here is that we unnecessarily install the puli cli in vendor. that seems pretty acceptable as worst case...

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,12 +13,12 @@
     "require": {
         "php": "^5.4|7.*",
         "puli/composer-plugin": "~1.0.0-beta8",
-        "puli/discovery": "~1.0.0-beta8"
+        "puli/discovery": "~1.0.0-beta8",
+        "puli/cli": "~1.0.0-beta9"
     },
     "require-dev": {
         "php-http/httplug": "1.0.0-beta",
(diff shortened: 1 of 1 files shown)


ddeboer:
:+1:

Nyholm:
I agree with this merge, but I want to state the implications of this. Since Puli/cli has dependencies on symfony/components: ~2.3 we can not use discovery in a Symfony 3.0 application.

sagikazarmark:
Should be fixed in a future version. Any suggestions?

ddeboer:
@Nyholm Yeah, fixed in [master](https://github.com/puli/cli/blob/master/composer.json).

Nyholm:
@ddroer yeah they just merged my PR. 

Yes. All dependencies must be stable until we can tag a stable release.

PR #40, opened 2016-01-04
adjust puli dependencies

dbu:


composer.json
```diff
@@ -13,8 +13,8 @@
     "require": {
         "php": "^5.4|7.*",
         "puli/composer-plugin": "~1.0.0-beta8",
-        "puli/discovery": "~1.0.0-beta8",
-        "puli/cli": "~1.0.0-beta9"
+        "puli/discovery": "1.0.0-beta8",
```
dbu: master is currently broken. we should be more flexible again when its fixed, as this will be a huge PITA to install if more than one repository depends on puli.

composer.json
```diff
@@ -13,8 +13,8 @@
...
+        "puli/cli": "~1.0.0-beta9@dev"
```
dbu: there is no tag with symfony 3 support yet.

sagikazarmark:
:+1: Will tag another version once this gets merged.

dbu:
i think we should merge this. for 0.6 we depend on puli/cli, stopping to do that we need to bump to 0.7 - if we want that. we discuss this in #41 and will change again. right now, this PR keeps the same functionality with the same drawbacks but actually working (i hope)

sagikazarmark:
:+1:

PR #47, opened 2016-01-15
Add puli.phar, remove cli dependency

sagikazarmark:


composer.json
```diff
@@ -36,14 +34,20 @@
         }
     },
     "scripts": {
-        "test": "vendor/bin/phpspec run",
-        "test-ci": "vendor/bin/phpspec run -c phpspec.yml.ci"
+        "test": "bin/phpspec run",
+        "test-ci": "bin/phpspec run -c phpspec.yml.ci"
     },
+    "bin": [
+        "bin/puli.phar"
```
dbu: ah, this is exposing puli.phar in the bin directory, right? is ./bin automatically in the path for the puli composer plugin? meaning nothing more needs to be done in libs using discovery? and once the puli plugin handles downloading the .phar as needed, we can simply remove this again without any effect?

if that is the case: great job, lets merge this and release a new version, with a note in the changelog that providing puli.phar is a temporary solution and people should not rely on it for having access to puli cli in their project.

sagikazarmark: Yes, this exposes puli.phar in the composer bin-dir (using a symlink). It is not necessarily ./bin, depends on the composer config. Yes, nothing more needs to be done, I just tested it.

I am still fighting with some HHVM issues.

The nice thing is that if you have puli globally installed, that version is used. This however can be bad as well. Older versions will freak out. But: #nice-job @webmozart

sagikazarmark:
It seems HHVM is not supported by Puli right now.

dbu:
then lets skip hhvm i guess?

sagikazarmark:
Yeah.

PR #50, opened 2016-02-06
Remove Puli composer dependency

sagikazarmark:


composer.json
```diff
@@ -11,16 +11,17 @@
...
     },
     "require-dev": {
         "php-http/httplug": "^1.0",
         "php-http/message-factory": "^1.0",
+        "puli/composer-plugin": "1.0.0-beta9",
         "phpspec/phpspec": "^2.4",
         "henrikbjorn/phpspec-code-coverage" : "^1.0"
     },
     "suggest": {
+        "puli/composer-plugin": "Allow to use Discovery",
```
dbu: this is weird. are the options to have the plugin globally installed or in composer.json? or does puli now download itself? but not the plugin i guess, that would need to be there in the first place...

my question is why would i include discovery in my project if not for using discovery... in what regard is the plugin optional?

sagikazarmark: You would include discovery to ensure that the classes provided by discovery are available. Although discovery will throw an exception if puli is not available. The idea is that Puli composer plugin is an environmental detail: you can install it in your application, if you want to use it. As discovery should always be a fallback, the fact that you have it installed does not mean that you actually use the discovery layer. This makes discovery optional (although our wrapper around Puli must be a hard dependency wherever it is used).

dbu: still fail to see the point, sorry. i could do
```
if (!$client) {
    if (!class_exists('Http\Discovery...')) {
        throw new Exception('Specify client or include discovery');
    }
    $client = ClientDiscovery::find();
}
```

making puli optional in discovery does not change anything here, except that including discovery in a project becomes more tricky.

sagikazarmark: As @webmozart pointed out, puli composer plugin is not necessary to use puli. You can manually define your bindings, etc as well.

Basically you want to set up puli in your application or as a dev dependency. If you don't, and a request is made to the discovery layer, it will look for a puli setup and throw an exception if there isn't any. This makes Puli configuration your responsibility, while discovery comes as a dependency.

The point is: Puli can be set up multiple ways and it should be the user's choice how to do it.

dbu: okay, now i got it. thank you for your patience. i agree with the change then!

dbu: can you improve the comment on the composer requirement however, so that other users (and my future self ;-) ) will be less confused? something like "If you want auto discovery to happen, use the puli composer plugin". maybe we could even link to the doc page where we (or puli) explains the other options that exist.

sagikazarmark: Yes, that description is not so good. The discovery page in the docs must also be improved as it is quite old.

joelwurtz:
:+1:

Issue #55, opened 2016-05-01
Make puli optional


sagikazarmark:
We had many complaints about using puli in the discovery layer. Although I love it and for me it made it easier to discover clients, not everyone feels the same.

One of the most common complaints is about dependencies: https://github.com/php-http/httplug/issues/101#issuecomment-176426964

I think Puli improved a lot on this front in the near past.

All in all: we should investigate the case and see if we can make Puli optional.

AFAIK @Nyholm came up with an idea (maybe something that Payum implemented?) to maintain a list of known resources in the discovery and fallback to Puli (if installed).

Althought I don't like the idea of having multiple discovery ways, I thought I would ask for feedback to see what other think about it.

Discovery was always something which I treated as a stepchildren. Personally, I don't like to use it. Makes things easier, but provides us less control.

dbu:
hm, indeed having autodiscovery for just guzzle and maybe 1 or 2 very
common other ways we know about and use puli for possibly unknown 3rd
party implementations seems like a good idea to me.

i would keep the order: check if we have puli, if we do use it. only
then do the simplified check. the other way round, there would be no way
to use puli if you happen to have one of our known implementations
laying around in the codebase

sagikazarmark:
Should it be in a new minor release?

It is not a patch, but doesn't break BC and many libs rely on ^0.8

dbu:
i think minor release is fine, as long as we keep using puli if its available, to not trip up existing applications that rely on puli.

Nyholm:
I like the idea of making Puli optional. It will make it easier to get started with Httplug. 
I've created a PR #58, but when I started to think about that PR I was not convinced that the solution was good enough. I created a second PR #59 which I like much more.

PR #58, opened 2016-06-10
Added fallback on common classes. Puli is now optional

Nyholm:
This PR will make Puli optional as discussed in #55. 

## This is how discovery works 
Step 1) Use puli,
 * If puli does not find anything then we are done. 
 * If Puli is not installed we continue to step 2.

Step 2) See if we find any HttpClients

Step 3) Look for Guzzle factories, will succeed if `php-http/message` and `guzzlehttp/psr7` is installed

Step 4) Look for Diactoros factories, will succeed if `php-http/message` and `zendframework/zend-diactoros` is installed

## Changes:
* When Puli is missing we will not throw a `\RuntimeException` anymore. There will be a `NotFoundException`.
* The `NotFoundException` is moved from `Http\Discovery` to `Http\Discovery\Exception`

## TODO
This is just a proof of concept. Do we like to go this way?

* [ ] Add tests
* [ ] Update documentation

### DIFF
--- a/spec/ClassDiscoverySpec.php
+++ b/spec/ClassDiscoverySpec.php
@@ -47,7 +47,7 @@ class ClassDiscoverySpec extends ObjectBehavior
     {
         $discovery->findBindings('InvalidBinding')->willReturn([]);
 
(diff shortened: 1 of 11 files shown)


Nyholm:
Closed in favor of #59

PR #59, opened 2016-06-12
Flexible discovery

Nyholm:
I was not too happy about #58. We treat Puli in a special manner which feels weird. Also, I was not too happy about the logic in the fallback strategies. 

This PR continues the work and replaces #58. The user has now the possibility to select what strategies to use and also add some more custom if there is a need for it. 

## Changes
* Introduced `Stratiegies` that do the discovery. We try the strategies one by one until we find what we're looking for
* many public functions are moved from `ClassDiscovery` to `Strategy\Puli`. 
* I've also created some new exceptions. 
* Added a simple cache mechanism
* When Puli is missing we will not throw a \RuntimeException anymore. There will be a NotFoundException which has previous exception so you can see that Puli was missing. 
* The NotFoundException is moved from Http\Discovery to Http\Discovery\Exception

## TODO
This is just a proof of concept. Do we like to go this way?

* [x] Add tests
* [ ] Update documentation

### DIFF
--- a/spec/ClassDiscoverySpec.php
+++ b/spec/ClassDiscoverySpec.php
@@ -3,29 +3,22 @@
 namespace spec\Http\Discovery;
 
 use Http\Discovery\ClassDiscovery;
+use Http\Discovery\Exception\DiscoveryFailedException;
+use Http\Discovery\Exception\NotFoundException;
+use Http\Discovery\Strategy\DiscoveryStrategy;
 use Puli\Discovery\Binding\ClassBinding;
 use Puli\GeneratedPuliFactory;
 use Puli\Discovery\Api\Discovery;
 use Puli\Repository\Api\ResourceRepository;
 use PhpSpec\ObjectBehavior;
+use spec\Http\Discovery\Helper\DiscoveryHelper;
 
 class ClassDiscoverySpec extends ObjectBehavior
 {
-    function let(
-        GeneratedPuliFactory $puliFactory,
-        ResourceRepository $repository,
-        Discovery $discovery
-    ) {
-        $puliFactory->createRepository()->willReturn($repository);
-        $puliFactory->createDiscovery($repository)->willReturn($discovery);
-
+    function let() {
+        ClassDiscovery::setStrategies([DiscoveryHelper::class]);
+        DiscoveryHelper::clearClasses();
         $this->beAnInstanceOf('spec\Http\Discovery\ClassDiscoveryStub');
-        $this->setPuliFactory($puliFactory);
-    }
-
-    function letgo()
-    {
-        $this->resetPuliFactory();
     }
 
     function it_is_initializable()
@@ -33,53 +26,35 @@ class ClassDiscoverySpec extends ObjectBehavior
         $this->shouldHaveType('Http\Discovery\ClassDiscovery');
     }
 
-    function it_has_a_puli_factory(GeneratedPuliFactory $puliFactory)
+    function it_throws_an_exception_when_no_candidate_found(DiscoveryStrategy $strategy)
     {
-        $this->getPuliFactory()->shouldReturn($puliFactory);
+        $strategy->getCandidates('NoCandidate')->willReturn([]);
+
+        $this->shouldThrow(DiscoveryFailedException::class)->duringFind('NoCandidate');
     }
 
-    function it_has_a_puli_discovery(Discovery $discovery)
+    function it_returns_a_class(DiscoveryStrategy $strategy)
     {
-        $this->getPuliDiscovery()->shouldReturn($discovery);
+        $candidate = ['class' => 'ClassName', 'condition' => true];
+        DiscoveryHelper::setClasses('Foobar', [$candidate]);
+
+        $this->find('Foobar')->shouldReturn('ClassName');
     }
 
-    function it_throws_an_exception_when_binding_not_found(Discovery $discovery)
-    {
-        $discovery->findBindings('InvalidBinding')->willReturn([]);
+    function it_validates_conditions(DiscoveryStrategy $strategy) {
+        $c0 = ['class' => 'ClassName0', 'condition' => false];
+        $c1 = ['class' => 'ClassName1', 'condition' => true];
+        $c2 = ['class' => 'ClassName2', 'condition' => false];
+        DiscoveryHelper::setClasses('Foobar', [$c0, $c1, $c2]);
 
-        $this->shouldThrow('Http\Discovery\NotFoundException')->duringFindOneByType('InvalidBinding');
-    }
-
-    function it_returns_a_class_binding(Discovery $discovery, ClassBinding $binding)
-    {
-        $binding->hasParameterValue('depends')->willReturn(false);
-        $binding->getClassName()->willReturn('ClassName');
-
-        $discovery->findBindings('Binding')->willReturn([$binding]);
-
-        $this->findOneByType('Binding')->shouldReturn('ClassName');
-    }
-
-    function it_returns_a_class_binding_with_dependency(
-        Discovery $discovery,
-        ClassBinding $binding1,
-        ClassBinding $binding2
-    ) {
-        $binding1->hasParameterValue('depends')->willReturn(true);
-        $binding1->getParameterValue('depends')->willReturn(false);
-
-        $binding2->hasParameterValue('depends')->willReturn(false);
-        $binding2->getClassName()->willReturn('ClassName');
-
-        $discovery->findBindings('Binding')->willReturn([
-            $binding1,
-            $binding2,
-        ]);
-
-        $this->findOneByType('Binding')->shouldReturn('ClassName');
+        $this->find('Foobar')->shouldReturn('ClassName1');
     }
 }
 
 class ClassDiscoveryStub extends ClassDiscovery
 {
+    public static function find($type)
+    {
+        return self::findOneByType($type);
+    }
 }
--- a/spec/Helper/DiscoveryHelper.php
+++ b/spec/Helper/DiscoveryHelper.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace spec\Http\Discovery\Helper;
+
+use Http\Discovery\Strategy\DiscoveryStrategy;
+
+/**
+ * This is a discovery helper used in tests
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class DiscoveryHelper implements DiscoveryStrategy
+{
+    private static $classes = [];
+
+    /**
+     * @param string $type
+     * @param array $classes
+     *
+     * @return $this
+     */
+    public static function setClasses($type, array $classes)
+    {
+        self::$classes[$type] = $classes;
+    }
+
+    /**
+     * Clear all classes.
+     */
+    public static function clearClasses()
+    {
+        self::$classes = [];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function getCandidates($type)
+    {
+        if (isset(static::$classes[$type])) {
+            return static::$classes[$type];
+        }
+
+        return [];
+    }
+}
--- a/spec/HttpAsyncClientDiscoverySpec.php
+++ b/spec/HttpAsyncClientDiscoverySpec.php
@@ -2,28 +2,24 @@
 
 namespace spec\Http\Discovery;
 
+use Http\Client\HttpAsyncClient;
+use Http\Discovery\ClassDiscovery;
+use Http\Discovery\Exception\NotFoundException;
+use Http\Discovery\Strategy\DiscoveryStrategy;
+use Prophecy\Argument;
 use Puli\GeneratedPuliFactory;
 use Puli\Discovery\Api\Discovery;
 use Puli\Discovery\Binding\ClassBinding;
 use Puli\Repository\Api\ResourceRepository;
 use PhpSpec\ObjectBehavior;
+use spec\Http\Discovery\Helper\DiscoveryHelper;
 
 class HttpAsyncClientDiscoverySpec extends ObjectBehavior
 {
-    function let(
-        GeneratedPuliFactory $puliFactory,
-        ResourceRepository $repository,
-        Discovery $discovery
-    ) {
-        $puliFactory->createRepository()->willReturn($repository);
-        $puliFactory->createDiscovery($repository)->willReturn($discovery);
-
-        $this->setPuliFactory($puliFactory);
-    }
-
-    function letgo()
+    function let()
     {
-        $this->resetPuliFactory();
+        ClassDiscovery::setStrategies([DiscoveryHelper::class]);
+        DiscoveryHelper::clearClasses();
     }
 
     function it_is_initializable()
@@ -36,15 +32,15 @@ class HttpAsyncClientDiscoverySpec extends ObjectBehavior
         $this->shouldHaveType('Http\Discovery\ClassDiscovery');
     }
 
-    function it_finds_an_async_http_client(
-        Discovery $discovery,
-        ClassBinding $binding
-    ) {
-        $binding->hasParameterValue('depends')->willReturn(false);
-        $binding->getClassName()->willReturn('spec\Http\Discovery\Stub\HttpAsyncClientStub');
+    function it_finds_an_async_http_client(DiscoveryStrategy $strategy) {
 
-        $discovery->findBindings('Http\Client\HttpAsyncClient')->willReturn([$binding]);
+        $candidate = ['class' => 'spec\Http\Discovery\Stub\HttpAsyncClientStub', 'condition' => true];
+        DiscoveryHelper::setClasses(HttpAsyncClient::class, [$candidate]);
 
         $this->find()->shouldImplement('Http\Client\HttpAsyncClient');
+    }
+
+    function it_throw_exception(DiscoveryStrategy $strategy) {
+        $this->shouldThrow(NotFoundException::class)->duringFind();
     }
 }
--- a/spec/MessageFactoryDiscoverySpec.php
+++ b/spec/MessageFactoryDiscoverySpec.php
@@ -2,28 +2,23 @@
 
 namespace spec\Http\Discovery;
 
+use Http\Discovery\ClassDiscovery;
+use Http\Discovery\Exception\NotFoundException;
+use Http\Discovery\Strategy\DiscoveryStrategy;
+use Http\Message\MessageFactory;
 use Puli\GeneratedPuliFactory;
 use Puli\Discovery\Api\Discovery;
 use Puli\Discovery\Binding\ClassBinding;
 use Puli\Repository\Api\ResourceRepository;
 use PhpSpec\ObjectBehavior;
+use spec\Http\Discovery\Helper\DiscoveryHelper;
 
 class MessageFactoryDiscoverySpec extends ObjectBehavior
 {
-    function let(
-        GeneratedPuliFactory $puliFactory,
-        ResourceRepository $repository,
-        Discovery $discovery
-    ) {
-        $puliFactory->createRepository()->willReturn($repository);
-        $puliFactory->createDiscovery($repository)->willReturn($discovery);
-
-        $this->setPuliFactory($puliFactory);
-    }
-
-    function letgo()
+    function let()
     {
-        $this->resetPuliFactory();
+        ClassDiscovery::setStrategies([DiscoveryHelper::class]);
+        DiscoveryHelper::clearClasses();
     }
 
     function it_is_initializable()
@@ -36,15 +31,17 @@ class MessageFactoryDiscoverySpec extends ObjectBehavior
         $this->shouldHaveType('Http\Discovery\ClassDiscovery');
     }
 
-    function it_finds_a_message_factory(
-        Discovery $discovery,
-        ClassBinding $binding
-    ) {
-        $binding->hasParameterValue('depends')->willReturn(false);
-        $binding->getClassName()->willReturn('spec\Http\Discovery\Stub\MessageFactoryStub');
+    function it_finds_a_message_factory(DiscoveryStrategy $strategy) {
 
-        $discovery->findBindings('Http\Message\MessageFactory')->willReturn([$binding]);
+        $candidate = ['class' => 'spec\Http\Discovery\Stub\MessageFactoryStub', 'condition' => true];
+        DiscoveryHelper::setClasses(MessageFactory::class, [$candidate]);
 
         $this->find()->shouldImplement('Http\Message\MessageFactory');
+    }
+
+    function it_throw_exception(DiscoveryStrategy $strategy) {
+        $strategy->getCandidates('Http\Message\MessageFactory')->willReturn([]);
+
+        $this->shouldThrow(NotFoundException::class)->duringFind();
     }
 }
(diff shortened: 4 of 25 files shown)


src/ClassDiscovery.php
```diff
@@ -2,126 +2,104 @@
...
+            return $class;
         }
 
-        self::$puliFactory = $puliFactory;
-        self::$puliDiscovery = null;
-    }
+        $exceptions = [];
+        foreach (self::$strategies as $strategy) {
+            try {
+                $bindings = call_user_func($strategy.'::find', $type);
```
Nyholm: Is "binding" the best name here?

src/Strategy/DiscoveryStrategy.php
```diff
@@ -0,0 +1,23 @@
...
+     * Find a resource of a specific type.
+     *
+     * @param string $type
+     *
+     * @return array The return value is always an array with zero or more elements. Each
+     *               element is an array with two keys ['class' => string, 'condition' => mixed].
+     *
+     * @throws StrategyUnavailableException if we cannot use this strategy.
+     */
+    public static function find($type);
```
Nyholm: Maybe not use "find" here because it is too similar to to the `ClassDiscovery`. To avoid confusion this maybe should be renamed to "getBinding" or "getClasses".

src/ClassDiscovery.php
```diff
@@ -2,126 +2,104 @@
...
      *
-     * @param object $puliFactory
+     * @param string $type
+     *
+     * @return string
+     *
+     * @throws DiscoveryFailedException
      */
-    public static function setPuliFactory($puliFactory)
+    protected static function findOneByType($type)
```
Nyholm: Made this function protected simply because I can't find any reason for it to be public.

sagikazarmark:
Huh, nice work. Haven't reviewed yet, but definitely will.

About strategies: not sure if we want discoveries to be made configurable. It's like creating a custom configuration layer which we don't want to.

I think the idea in the linked ticket was to use puli if available, fallback to our known resources if they are available. If this is what strategies intend to do (with a better abstraction) then it's fine, but we should keep it minimal: not too much extension points, etc

Nyholm:
Thank you Mark. 

>I think the idea in the linked ticket was to use puli if available, fallback to our known resources if they are available.

Have a look at #58. It did that exactly. But it felt weird and it did not follow the Open/Close principle. But maybe you are correct, this is too many of extension points.

sagikazarmark:
Discovery is for convenience, so I would rather follow KISS.

spec/ClassDiscoverySpec.php
```diff
@@ -47,7 +47,7 @@ function it_throws_an_exception_when_binding_not_found(Discovery $discovery)
     {
         $discovery->findBindings('InvalidBinding')->willReturn([]);
 
-        $this->shouldThrow('Http\Discovery\NotFoundException')->duringFindOneByType('InvalidBinding');
+        $this->shouldThrow('Http\Discovery\Exception\NotFoundException')->duringFindOneByType('InvalidBinding');
```
dbu: should we use NotFoundException::class here?

src/ClassDiscovery.php
```diff
@@ -2,126 +2,104 @@
...
-     * Resets the factory.
-     */
-    public static function resetPuliFactory()
-    {
-        self::$puliFactory = null;
-        self::$puliDiscovery = null;
-    }
+            foreach ($bindings as $binding) {
+                if ($binding['condition']) {
+                    if (!self::evaluateCondition($binding['condition'])) {
```
dbu: are these conditions static? does the cache indeed not need to check conditions again?

src/Strategy/DiscoveryStrategy.php
dbu: getCandidates maybe? because there is also the condition to take care of.

src/Strategy/AbstractStrategy.php
```diff
@@ -0,0 +1,23 @@
...
+{
+    protected static $classes = [];
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function find($type)
+    {
+        if (isset(static::$classes[$type])) {
+            return static::$classes[$type];
```
dbu: should this check if the class indeed exists? otherwise the concrete strategies would need to do that check.

src/Strategy/Puli.php
```diff
@@ -0,0 +1,108 @@
...
+ */
+class Puli implements DiscoveryStrategy
+{
+    /**
+     * @var \Puli\GeneratedPuliFactory
+     */
+    private static $puliFactory;
+
+    /**
+     * @var \Puli\Discovery\Api\Discovery
```
dbu: use statements?

src/Strategy/Puli.php
```diff
@@ -0,0 +1,108 @@
...
+     */
+    private static $puliDiscovery;
+
+    /**
+     * @return \Puli\GeneratedPuliFactory
+     *
+     * @throws PuliUnavailableException
+     */
+    public static function getPuliFactory()
```
dbu: why are these methods public? imho private would be enough for them. except the setter.

src/Strategy/Puli.php
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        self::$puliFactory = $puliFactory;
+        self::$puliDiscovery = null;
+    }
+
+    /**
+     * Resets the factory.
+     */
+    public static function resetPuliFactory()
```
dbu: do we need this? it feels fishy

dbu:
i like it. its indeed quite a thing you built here, but i think its clearly structured. i think there are too many public methods that are additional extension points we should not need. i think hard-setting the answer for a $type on the discovery itself should be all the configuration we allow. the whole discovery interface and implementations should have `@internal` annotations and never be interacted with. if you want to interact, have your application/library accept instantiated clients in addition to the discovery.

i also feel a bit unsure about the number of static methods. but as we don't want the discoveries to be configurable anyways, it should not matter.

src/ClassDiscovery.php
sagikazarmark: My idea was that composer plugin is not mandatory, so the constant might not always be available. Not sure if that makes sense though.

src/ClassDiscovery.php
sagikazarmark: Puli terminology

src/ClassDiscovery.php
Nyholm: I assumed (maybe wrongly) that they are static. 

The current cache implementation is only valid for a request and our conditions are checking if the class exists. But since we support conditions being a callable we may need to rerun the conditions to be sure.

src/Strategy/AbstractStrategy.php
sagikazarmark: If strategies dont do the test, they dont know if the current resource is available and cant proceed to the next one.

Nyholm: The strategies only finds classes and their conditions. It is the ClassDiscovery that validates if the conditions are met and decides if the class should be used or not.

src/Strategy/Puli.php
Nyholm: They were public before. I guess (I do not know) they are public because it would make the tests simpler. I'll change them to private.

src/ClassDiscovery.php
Nyholm: Yeah, but Im hesitant over using "binding" since this is not in the Puli scope. I'll use "binding"

src/Strategy/AbstractStrategy.php
Nyholm: > If strategies dont do the test, they dont know if the current resource is available and cant proceed to the next one.

Im not sure that is true anymore. Each strategy return all *candidates* they know of. It is the `ClassDiscovery` that makes a decision if we should use a candidate and if we should visit the next strategy.

sagikazarmark: Ah, okay, missed that.

Nyholm:
FYI, If Puli is not installed **AND** we can't find a client: 

For a client:
image

For a message factory: 
image

src/Strategy/Puli.php
Nyholm: I did not add use statements for one (silly) reason: When Puli is not installed my IDE will mark this class "with errors" because the class in the use statement does not exist.

I'll add them if anyone wants them.

dbu: i think my IDE also complains about missing classes in annotations...
for the language php, its ok to have use statements for things that
don't exist. its even ok to have typehints as long as you default to
null and only pass null...

i'd prefer the use, to be consistent.

dbu:
i like this! the implementation looks good to me.

src/Strategy/HttpClients.php
```diff
@@ -0,0 +1,38 @@
...
+            ['class' => Curl::class, 'condition' => Curl::class],
+            ['class' => React::class, 'condition' => React::class],
+        ],
+        'Http\Client\HttpClient' => [
+            ['class' => Guzzle6::class, 'condition' => Guzzle6::class],
+            ['class' => Guzzle5::class, 'condition' => Guzzle5::class],
+            ['class' => Curl::class, 'condition' => Curl::class],
+            ['class' => Socket::class, 'condition' => Socket::class],
+            ['class' => Buzz::class, 'condition' => Buzz::class],
+            ['class' => React::class, 'condition' => React::class],
```
joelwurtz: Not sure if this work when you have only one HttpClient and not the other since you will only have one class, the other will not be found and this will throw an autoloading exception.

Nyholm: That is the beauty of Foo::class, it doesn't throw an exception if the class not exists. =)

dbu:
@Nyholm can you check the build failures please?

Nyholm:
Yes of course. I started to fix the tests this morning but did not have time to finish.

dbu:
sure, no hurry. just wanted to be sure we understand who is waiting on whom ;-)

Nyholm:
Yooho. Tests are passing. I added a fix for the [composer/puli thing](https://github.com/puli/issues/issues/190). 

Please feel free to make suggestions on the tests. PHPSpec is not my strongest suit.

spec/Helper/DiscoveryHelper.php
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    /**
+     * Clear all classes.
+     */
+    public static function clearClasses()
+    {
+        self::$classes = [];
+    }
+}
```
sagikazarmark: Missing newline

spec/HttpAsyncClientDiscoverySpec.php
```diff
@@ -36,15 +32,15 @@ function it_is_a_class_discovery()
...
+        DiscoveryHelper::setClasses('Http\Client\HttpAsyncClient', [$candidate]);
 
         $this->find()->shouldImplement('Http\Client\HttpAsyncClient');
     }
-}
+
+    function it_throw_exception(DiscoveryStrategy $strategy) {
+        $this->shouldThrow(NotFoundException::class)->duringFind();
+    }
+}
```
sagikazarmark: Missing newline

spec/Strategy/PuliSpec.php
```diff
@@ -0,0 +1,108 @@
...
+    }
+    /**
+     * Resets the factory.
+     */
+    public static function resetPuliFactory()
+    {
+        self::$puliFactory = null;
+        self::$puliDiscovery = null;
+    }
+}
```
sagikazarmark: Missing newline

spec/HttpAsyncClientDiscoverySpec.php
```diff
@@ -36,15 +32,15 @@ function it_is_a_class_discovery()
...
-        Discovery $discovery,
-        ClassBinding $binding
-    ) {
-        $binding->hasParameterValue('depends')->willReturn(false);
-        $binding->getClassName()->willReturn('spec\Http\Discovery\Stub\HttpAsyncClientStub');
+    function it_finds_an_async_http_client(DiscoveryStrategy $strategy) {
 
-        $discovery->findBindings('Http\Client\HttpAsyncClient')->willReturn([$binding]);
+        $candidate = ['class' => 'spec\Http\Discovery\Stub\HttpAsyncClientStub', 'condition' => true];
+        DiscoveryHelper::setClasses('Http\Client\HttpAsyncClient', [$candidate]);
```
sagikazarmark: I think we can use the class constant here too

spec/MessageFactoryDiscoverySpec.php
```diff
@@ -2,30 +2,25 @@
...
-    function letgo()
+    function let()
     {
-        $this->resetPuliFactory();
+        ClassDiscovery::clearCache();
+        ClassDiscovery::$strategies = [DiscoveryHelper::class];
+        DiscoveryHelper::clearClasses();
     }
-
+
```
sagikazarmark: Extra indent?

src/ClassDiscovery.php
```diff
@@ -2,126 +2,117 @@
...
+    protected static function findOneByType($type)
     {
-        if (null === self::$puliFactory) {
-            if (!defined('PULI_FACTORY_CLASS')) {
-                throw new \RuntimeException('Puli Factory is not available');
-            }
-
-            $puliFactoryClass = PULI_FACTORY_CLASS;
+        // Look in the cache
+        if (null !== $class = self::getFromCache($type)) {
```
sagikazarmark: Although not necessary, I would use this: `null !== ($class = self::getFromCache($type))` More readable I think

sagikazarmark:
Let me see if I understand well: we have x strategies which all return some resources, which we cache and then use this cache to find the actual resource?

Nyholm:
Thank you for your review. 

We work with a strategy at the time, if a strategy return a "candidate" were its dependencies are met we have found a class and we will return without investigating further strategies. This candidate is saved to the cache. The cache is only read from the second time we ask for the same resource. 

The great thing about the strategies is that we can easily support Puli2 or WhatveverDiscovery without editing any existing class.

src/ClassDiscovery.php
```diff
@@ -2,126 +2,117 @@
...
 abstract class ClassDiscovery
 {
     /**
-     * @var GeneratedPuliFactory
+     * A list of strategies to find classes.
+     *
+     * @var array
      */
-    private static $puliFactory;
+    public static $strategies = [
```
sagikazarmark: Do we need to configure the strategies in the abstract class? Can't we do it on a per discovery basis?

src/ClassDiscovery.php
```diff
@@ -2,126 +2,117 @@
...
+        Strategy\Puli::class,
+        Strategy\HttpClients::class,
```
sagikazarmark: Actually I wonder if we should make these three (HttpClients, Guzzle, Diactoros) to be one strategy. Since discovery is done based on type, I see no harm in having them in one array. If this is possible, it invalidates my comment a few lines above

src/ClassDiscovery.php
```diff
@@ -2,126 +2,117 @@
...
-        if (!isset(self::$puliDiscovery)) {
-            $factory = self::getPuliFactory();
-            $repository = $factory->createRepository();
+        if (!isset(self::$cache[$type])) {
+            return;
+        }
 
-            self::$puliDiscovery = $factory->createDiscovery($repository);
+        $candidate = self::$cache[$type];
+        if (!self::evaluateCondition($candidate['condition'])) {
```
sagikazarmark: Why do we need to evaluate the condition here? We already did that, otherwise the given candidate would not be cached at all. Or would it?

src/ClassDiscovery.php
```diff
@@ -2,126 +2,117 @@
...
+            return;
         }
 
-        return self::$puliDiscovery;
+        return $candidate['class'];
```
sagikazarmark: Do we need to cache the condition then? Maybe it's enough to cache only the class?

src/ClassDiscovery.php
```diff
@@ -2,126 +2,117 @@
...
     }
 
     /**
-     * Finds a resource.
-     *
-     * @return object
+     * Clear the cache.
      */
-    public static function find()
+    public static function clearCache()
```
sagikazarmark: I wonder if should only clear the cache of a given type. When called in a specific discovery, the cache for all types will be wiped.

src/Exception/DiscoveryFailedException.php
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Thrown when all discovery strategies fails to find a resource.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+final class DiscoveryFailedException extends \RuntimeException
+{
+    /**
+     * @var array
```
sagikazarmark: I guess this is \Exception[] then?

src/Exception/DiscoveryFailedException.php
```diff
@@ -0,0 +1,34 @@
...
+    private $exceptions;
+
+    /**
+     * @param $exceptions
+     */
+    public function __construct($message, array $exceptions = [])
+    {
+        $this->exceptions = $exceptions;
+
+        parent::__construct($message, 0, array_shift($exceptions));
```
sagikazarmark: Hm, does previous exception make sense here?

src/HttpAsyncClientDiscovery.php
```diff
@@ -15,11 +17,21 @@
...
      *
      * @return HttpAsyncClient
+     *
+     * @throws NotFoundException
      */
     public static function find()
     {
-        $asyncClient = static::findOneByType('Http\Client\HttpAsyncClient');
+        try {
+            $asyncClient = static::findOneByType('Http\Client\HttpAsyncClient');
```
sagikazarmark: We could use the class constant here

src/HttpAsyncClientDiscovery.php
```diff
@@ -15,11 +17,21 @@
...
 
-        return new $asyncClient();
+            return new $asyncClient();
+        } catch (DiscoveryFailedException $e) {
+            throw new NotFoundException(
+                'No Httplug async clients found. Make sure to install a package providing "php-http/async-client-implementation". Example: "php-http/guzzle6-adapter".',
```
sagikazarmark: Just for consistency: HTTPlug would be better.

src/Strategy/AbstractStrategy.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace Http\Discovery\Strategy;
+
+/**
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+abstract class AbstractStrategy implements DiscoveryStrategy
```
sagikazarmark: If we merge the static strategies into one, maybe this becomes obsolete?

src/StreamFactoryDiscovery.php
```diff
@@ -15,14 +17,16 @@
...
+     * @throws NotFoundException
      */
     public static function find()
     {
         try {
             $streamFactory = static::findOneByType('Http\Message\StreamFactory');
 
             return new $streamFactory();
-        } catch (NotFoundException $e) {
+        } catch (DiscoveryFailedException $e) {
```
sagikazarmark: Maybe say what factory?

src/StreamFactoryDiscovery.php
```diff
@@ -15,14 +17,16 @@
...
             throw new NotFoundException(
```
sagikazarmark: Maybe say what factory?

sagikazarmark:
This is cool, I like it, great job @Nyholm.

I've added a few more comments besides the CS ones from earlier. Can you please take a look at them and at least answer them? If you have no time, I can do the improvements I suggested, but I need your comment to make sure I understood everything right. Thank you!

src/ClassDiscovery.php
Nyholm: No, we do not **need** to have those as separate strategies. We could merge them, I had them separate because it looked good. (Highly opinionated in other words). I'll merge them because of https://github.com/php-http/discovery/pull/59#discussion_r68453992

src/ClassDiscovery.php
Nyholm: Yes, we are using strategies that we know would fail. Example; the HttpClients strategy will always fail for the FactoryDiscovery. I'll fix this by merging the strategies suggested bellow.

src/ClassDiscovery.php
Nyholm: @dbu pointed out (https://github.com/php-http/discovery/pull/59#discussion_r66913507) that conditions are not necessarily static, so the conditions have to be revalidated.

src/ClassDiscovery.php
Nyholm: We do because conditions are not static. https://github.com/php-http/discovery/pull/59#discussion_r66913507

src/ClassDiscovery.php
sagikazarmark: Fair enough

src/ClassDiscovery.php
Nyholm: Good, I should move this to the Discovery classes? Though, I created this because I had to make the tests work. The only real use case I can see is where you use discovery and then add strategies yourself. Then you have to clear the cache to make the new strategies to work.

src/Exception/DiscoveryFailedException.php
Nyholm: Yeah, sorry.

src/ClassDiscovery.php
sagikazarmark: As I can see one can add a strategy by setting a public property. Not sure if it is a good idea since then you have to manually clear the cache. I think it would be better to have a method for adding strategies (or replacing the whole list), so that cache is automatically cleared. You can use it as a "hack" in tests to manually clear the cache.

src/Exception/DiscoveryFailedException.php
Nyholm: This is a hack to make sure we at least get one exception in our exception change. Without this you will never see the "Puli is not installed"- exception.

src/HttpAsyncClientDiscovery.php
Nyholm: Sure, I had that that in mind as well.

src/Exception/DiscoveryFailedException.php
sagikazarmark: I see. Well, I don't have a better alternative.

src/HttpAsyncClientDiscovery.php
Nyholm: Of course

src/Strategy/AbstractStrategy.php
Nyholm: Agreed

Nyholm:
Thank you. I've address all your comments, either with a reply or a code change (or both).

sagikazarmark:
This is still an open question from my POV: https://github.com/php-http/discovery/pull/59#discussion_r68460108

src/ClassDiscovery.php
Nyholm: Agreed. Thank you. I've updated the PR.

src/StreamFactoryDiscovery.php
Nyholm: Fixed

sagikazarmark:
Great, I am happy with the current state. Can you please add some change log and squash?

Nyholm:
Awesome. I'll must go afk now. Please feel free to make some changes or merge this when the tests passes. 

Thank you for the reviews!

PR #60, opened 2016-06-27
We are using ::class which is php 5.5

Nyholm:


sagikazarmark:
Ah, okay. Clear now.

xabbuh:
Was it possible before to use some code from the package even with PHP 5.4? Because in that case it might be a good idea to add an upgrade file to clearly document this change.

Nyholm:
If you only using the puli strategy and not running tests, then you were able to use php 5,4 (I think)

Yes. We should document that. I'll might have time to do it later tonight. 

For the record: PHP 5.5 is EOL within a couple of weeks (days?). We should not feel bad for dropping 5.4 support.

dbu:
i agree with dropping 5.4. but the changelog entry is a good idea in
case somebody is wondering what is going on. afaik many of our parts are
5.5 already.

Issue #57, opened 2016-06-03
Find async or non-async client


sagikazarmark:
In some cases it doesn't matter if the installed client is sync or async (for example when known to be used in PluginClient). For that we could have a discovery which finds a sync or async client.

dbu:
or the code could just try to discover whatever it prefers, catch the
exception if none found and try the other. i would not overdo it with
factories ;-)

sagikazarmark:
Actually, I was thinking about the same :smile: 

But it would still be useful as a snippet somewhere.

dbu:
sure, mentioning it in the doc would make a lot of sense. maybe a
section in the client discovery doc?

joelwurtz:
Or we could add an argument to the find method where we allow emulated client

```
HttpAsyncClientDiscovery::find($allowEmulation = false);
```

So if no async client is found and allowEmulation is set to true it use the Sync Client with an emulated async.

WDYT ?

Nyholm:
The only time you use `HttpAsyncClientDiscovery::find()` is when you are a library author. If you are a application author and writing your own code then you probably do not use discovery at all. 

As a library author you usually do not care if the client is emulated or not, you just want the async client api. 

The only time you care is if you are an application author using a library that is using `HttpAsyncClientDiscovery::find()`. Then you want to make sure discovery will find an async client. With that said, we cant have any special parameters to `find()` nor can you have a special `HttpAsyncClientDiscovery::findNoEmulated()`. 

What I propose is that we, by default, are trying to find all real async clients and then fallback on emulated ones.

sagikazarmark:
I would rather keep it simple and let the developer handle the situation. We don't know the preference of a given code. Adding emulation related code to the discovery adds an extra dependency (client-common) and complexity.

dbu:
i agree that we should keep discovery simple. you can try to find async,
if not found find sync and wrap it, if that is what you need. finding a
sync client and hiding that by wrapping it in an emulator is doing a
disservice, and adding tons of configuration options is complicating
things too much imho.

if somebody really wants to use discovery and the emulated client (e.g.
using a library that only finds an async client and has no DI, but you
need to use it with a sync client), they can still explicitly set the
emulated client on the async discovery. so the current code does cover
all conceivable situations, but makes emulation a bit more effort for
the sake of transparency and simplicity

Nyholm:
Okey. Big -1 for adding more dependency. 

[...]

PR #65, opened 2016-06-30
POC how an emulated async strategy may look like

Nyholm:
This PR should not be merged. I just want to show how I, as an application author, may force discovery to emulate an async client. 

Ref: https://github.com/php-http/discovery/issues/57

### DIFF
--- a/src/ClassDiscovery.php
+++ b/src/ClassDiscovery.php
@@ -36,7 +36,7 @@ abstract class ClassDiscovery
      *
      * @param string $type
      *
-     * @return string
+     * @return string|\Closure
      *
      * @throws DiscoveryFailedException
      */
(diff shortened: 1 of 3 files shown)


sagikazarmark:
I don't really like this. It adds an optional dependency. And IMO this is not how resources should be resolved.

Also, this is not much code: I would just have something similar in the docs or cookbook.

Nyholm:
Sorry if I wasn't clear. The EmulateAsyncClientStrategy does not belong in this repo for the reason you mention. But if we should allow a class similar to EmulateAsyncClientStrategy in the cookbook we have to make the changes in the ClassDiscovery and the the other *Discovery classes. 

I just want your feedback/opinions/suggestions for the changes in the ClassDiscovery and the *Discovery classes.

src/ClassDiscovery.php
```diff
@@ -36,7 +36,7 @@
      *
      * @param string $type
      *
-     * @return string
+     * @return string|\Closure
```
Nyholm: Do we want to allow this function to return a Closure?

Nyholm:
Im closing this PR in favor for #66.

PR #66, opened 2016-07-02
Allow for a candidate to have a Closure as a class parameter

Nyholm:
This PR will allow you to return a closure as a class parameter on the candidate. The good thing about this is that someone can write their own strategy that emulates a async client. An example of such strategy is seen in #65. 

This will increase the extendibility in the discovery mechanism.

### DIFF
--- a/src/ClassDiscovery.php
+++ b/src/ClassDiscovery.php
@@ -2,6 +2,7 @@
 
 namespace Http\Discovery;
 
+use Http\Discovery\Exception\ClassinstantiationFailedException;
 use Http\Discovery\Exception\DiscoveryFailedException;
 use Http\Discovery\Exception\StrategyUnavailableException;
 
@@ -36,7 +37,7 @@ abstract class ClassDiscovery
      *
      * @param string $type
      *
-     * @return string
+     * @return string|\Closure
      *
      * @throws DiscoveryFailedException
      */
@@ -176,5 +177,29 @@ abstract class ClassDiscovery
         }
 
         return false;
+    }
+
+    /**
+     * Get an instance of the $class.
+     *
+     * @param string|\Closure $class A FQN of a class or a closure that instantiate the class.
+     *
+     * @return object
+     */
+    protected static function instantiateClass($class)
+    {
+        try {
+            if (is_string($class)) {
+                return new $class();
+            }
+
+            if (is_callable($class)) {
+                return $class();
+            }
+        } catch (\Exception $e) {
+            throw new ClassinstantiationFailedException('Unexcepced exception when instantiating class.', 0, $e);
+        }
+        
+        throw new ClassinstantiationFailedException('Could not instantiate class becuase parameter is neitehr a callable or a string');
     }
 }
--- a/src/ClassDiscovery.php
+++ b/src/ClassDiscovery.php
@@ -199,7 +199,7 @@ abstract class ClassDiscovery
         } catch (\Exception $e) {
             throw new ClassinstantiationFailedException('Unexcepced exception when instantiating class.', 0, $e);
         }
-        
+
         throw new ClassinstantiationFailedException('Could not instantiate class becuase parameter is neitehr a callable or a string');
     }
 }
--- a/src/Exception/ClassinstantiationFailedException.php
+++ b/src/Exception/ClassinstantiationFailedException.php
@@ -2,11 +2,13 @@
 
 namespace Http\Discovery\Exception;
 
+use Http\Discovery\Exception;
+
 /**
  * Thrown when a class fails to instantiate.
  *
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
  */
-class ClassinstantiationFailedException extends \RuntimeException
+class ClassinstantiationFailedException extends \RuntimeException implements Exception
 {
 }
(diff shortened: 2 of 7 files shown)


dbu:
squash and merge?

Nyholm:
I rebased the PR and added `implements Exception`. 

FYI. I do not want the ClassDiscovery to instantiate the class because the `HttpClientDiscovery` may want to give a parameter to the closure in the future. Im not sure that will ever happen, but at least this will give room for that possibility.

dbu:
thanks!

Nyholm:
Thank you for merging.

src/ClassDiscovery.php
```diff
@@ -177,4 +178,28 @@ protected static function evaluateCondition($condition)
 
         return false;
     }
+
+    /**
+     * Get an instance of the $class.
+     *
+     * @param string|\Closure $class A FQN of a class or a closure that instantiate the class.
```
sagikazarmark: FQCN

src/ClassDiscovery.php
```diff
@@ -177,4 +178,28 @@ protected static function evaluateCondition($condition)
...
+            }
+
+            if (is_callable($class)) {
+                return $class();
+            }
+        } catch (\Exception $e) {
+            throw new ClassinstantiationFailedException('Unexcepced exception when instantiating class.', 0, $e);
+        }
+
+        throw new ClassinstantiationFailedException('Could not instantiate class becuase parameter is neitehr a callable or a string');
```
sagikazarmark: s/neitehr/neither

sagikazarmark: neither ... nor ...

src/Exception/ClassinstantiationFailedException.php
```diff
@@ -0,0 +1,14 @@
...
+namespace Http\Discovery\Exception;
+
+use Http\Discovery\Exception;
+
+/**
+ * Thrown when a class fails to instantiate.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class ClassinstantiationFailedException extends \RuntimeException implements Exception
```
sagikazarmark: Should this be final?

src/HttpAsyncClientDiscovery.php
```diff
@@ -24,7 +24,7 @@ public static function find()
         try {
             $asyncClient = static::findOneByType(HttpAsyncClient::class);
 
-            return new $asyncClient();
+            return static::instantiateClass($asyncClient);
```
sagikazarmark: Do we need the instantiation within the try? Is it possible that a closure throws a DiscoveryFailedException?

sagikazarmark:
Added a few comments. Nice work. :+1:

src/HttpAsyncClientDiscovery.php
Nyholm: No, we do not need a try-catch here. We know that instantiateClass could only throw a `ClassInstantiationFailedException`

src/Exception/ClassinstantiationFailedException.php
Nyholm: Yes, and also a capital "I" in Instantiation

src/HttpAsyncClientDiscovery.php
sagikazarmark: Then maybe we should move the instantiation out of the try-catch block. Makes the code easier to read IMO.

Nyholm: Sorry, I think I miss-read your comment. I meant that we do not need an *additional* try-catch here (at this line). 

> Do we need the instantiation within the try? Is it possible that a closure throws a DiscoveryFailedException?

Yes, we need the instantiation within a try-catch. Any weird exception could happen both with a FQCN (Class not found) and with the closure. The closure may throw a "VeryRandomException" that is not part of our code.

dbu: > Yes, we need the instantiation within a try-catch. Any weird exception
> could happen both with a FQCN (Class not found) and with the closure.
> The closure may throw a "VeryRandomException" that is not part of our code.

makes sense to me.

sagikazarmark: Correct me if I am wrong, but this try-catch only handles DiscoveryFailedExceptions, which mean we cannot (and we don't want to) catch VeryRandomExceptions.

Nyholm: Look here: https://github.com/php-http/discovery/blob/master/src/ClassDiscovery.php#L191-L206

Nyholm: Do you propose that we should rewrite it like: 
```php
        try {
            $asyncClient = static::findOneByType(HttpAsyncClient::class);
        } catch (DiscoveryFailedException $e) {
            throw new NotFoundException(
                'No HTTPlug async clients found. Make sure to install a package providing "php-http/async-client-implementation". Example: "php-http/guzzle6-adapter".',
                0,
                $e
            );
        }
        return static::instantiateClass($asyncClient);

```

sagikazarmark: I don't understand how the linked method is relevant: the exceptions there would still not be caught by this catch block.

Yes, I propose exactly that.

Nyholm: Now I understand you. You got a PR.

PR #67, opened 2016-07-02
Base exception

Nyholm:
Writing #66 made me think of introducing a base exception that all our exception extends from. This will make it easier to catch a discovery exception in case something goes wrong. 

This will be extra relevant if #66 gets merged because it introduces another exception.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### Added
 
 - Added back `Http\Discovery\NotFoundException` to preserve BC with 0.8 version
+- Added interface `Http\Discovery\Exception` which is implemented by all our exceptions
 
 ### Changed
 
--- a/src/Exception.php
+++ b/src/Exception.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace Http\Discovery;
+
+/**
+ * An interface implemented by all discovery related exceptions.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+interface Exception
+{
+}
--- a/src/Exception/NotFoundException.php
+++ b/src/Exception/NotFoundException.php
@@ -2,6 +2,8 @@
 
 namespace Http\Discovery\Exception;
 
+use Http\Discovery\Exception;
+
 /**
  * Thrown when a discovery does not find any matches.
  *
@@ -9,6 +11,6 @@ namespace Http\Discovery\Exception;
  *
  * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
  */
-/*final */class NotFoundException extends \RuntimeException
+/*final */class NotFoundException extends \RuntimeException implements Exception
 {
 }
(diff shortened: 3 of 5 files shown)


Nyholm:
But you cant catch an interface, right? Or can you? If so, an interface is way better. 

```php
try {
$client = HttpClientDiscovery::find()
} catch(DiscoveryExceptionInterface $e) {
 // log
}
```

sagikazarmark:
Of course you can. As long as the implementation extends `\Exception`. Leave the interface suffix. Just Http\Discovery\Exception. See the httplug repo.

Nyholm:
Cool, thank you. I've updated the PR.

src/Exception.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace Http\Discovery;
+
+/**
+ * An interface implemented by all our exception.
```
sagikazarmark: exceptionS

sagikazarmark:
Can you please add a changelog entry too?

Nyholm:
PR updated.

src/Exception.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace Http\Discovery;
+
+/**
+ * An interface implemented by all our exceptions.
```
dbu: s/by all our/by all discovery related/

src/Exception/NotFoundException.php
```diff
@@ -2,11 +2,13 @@
 
 namespace Http\Discovery\Exception;
 
+use Http\Discovery\Exception;
```
dbu: should the interface itself not live in the exception namespace?

Nyholm: I followed the example in HTTPlug: https://github.com/php-http/httplug/tree/master/src

I did some (very little) research and it seams that (a few) others are doing the same.

sagikazarmark: Exception\Exception sounds weird, also, the inheritance is also reflected in the naming this way, which seems to be logical to me.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 ### Changed
 
 - Puli strategy renamed to Puli Beta strategy to prevent incompatibility with a future Puli stable
+- Added interface `Http\Discovery\Exception` which is implemented by all our exceptions
```
sagikazarmark: Shouldn't this go into an Added section instead? Look for earlier examples.

Nyholm: Yes, I guess thats something that got lost when I rebased.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
 
 - Puli strategy renamed to Puli Beta strategy to prevent incompatibility with a future Puli stable
 
+### Added
```
sagikazarmark: Can you please put these in alphabetical order?

Nyholm:
Rebased and squashed.

Nyholm:
Thank you for merging.

PR #68, opened 2016-07-02
Add prepend and append strategy

Nyholm:
Im currently working on HTTPlug bundle and I will soon be in need for prepending a strategy to the existing ones. That is why I need a getter.

### DIFF
--- a/src/ClassDiscovery.php
+++ b/src/ClassDiscovery.php
@@ -4,6 +4,7 @@ namespace Http\Discovery;
 
 use Http\Discovery\Exception\DiscoveryFailedException;
 use Http\Discovery\Exception\StrategyUnavailableException;
+use Http\Discovery\Strategy\DiscoveryStrategy;
 
 /**
  * Registry that based find results on class existence.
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Why do you need a getter? Wondering about valid usecases, because strategies are internal thing, we shouldn't rely on them.

Nyholm:
I have an idea (work in progress). The HTTPlug's web toolbar shows all requests made by registered clients. But if a third party library uses discovery to find a client, that request will not be logged in the toolbar. 

My plan in to prepend the current list of strategies with a custom one made for HTTPlug. It would be a strategy of registered clients. I think I can make that work... Anyhow I need to do something like this: 

```php
$currentStrategies = HttpClientDiscovery::getStrategies();
$strategies = array_unshift($currentStrategies, CustomStrategy::class);
HttpClientDiscovery::setStrategies(strategies);
```

sagikazarmark:
For that, I would rather suggest something like `prependStrategy` method. (You can also add `appendStrategy`)

sagikazarmark:
Would you prefer to merge and release 0.9.x or you wait for 1.0 in the HttplugBundle PR?

Nyholm:
Let's wait a couple of days.. Ill reference this PR in a larger one in HttplugBundle

sagikazarmark:
Okay, let's wait with the tag, but is this ready? Can it be merged?

Nyholm:
Yes. I believe so.

Issue #62, opened 2016-06-29
Ready for stable release


Nyholm:
I started this issue from the discussion here: https://github.com/php-http/HttplugBundle/pull/81

Our BC promise in this library is the following functions:
```php
HttpAsyncClientDiscovery::find();
HttpClientDiscovery::find();
MessageFactoryDiscovery::find();
StreamFactoryDiscovery::find();
UriFactoryDiscovery::find();

ClassDiscovery::setStrategies();
ClassDiscovery::clearCache();

// And the
DiscoveryStrategy interface

// And our exceptions and when they are thrown
```

This API has nothing to do with the discovery methods (strategies) itself. The strategy classes should not be a part of our API promise. Which is a great thing. 

If we change the namespace of our clients we just add another "NewNamespacesStrategy" together with our existing `CommonClassesStrategy`.  Same with if Puli changes their API when they release 1.0. We just add a "Puli1Strategy". 

So if we leave the strategies outside our BC promise, are we ready for a stable release? At least, can we mature in to this API and tag 1.0 after the summer?

joelwurtz:
Was also thinking the same, as we have now a stable api for discovery it is good for a stable release, maybe we can rename Puli strategy to PuliBetaStrategy in the meantime ? (So we can create a PuliStrategy when it is stable)

dbu:
+1 for PuliBetaStrategy to make clear this is beta. once its stable, i
would hope that one strategy might work for all puli 1 versions, maybe
even puli 2...

sagikazarmark:
I would add one more thing here: NotFoundException. The API became quite stable in 0.8 and many libraries (like mailgun) started to use it. I would really hate if they haven't updated because of the BC break between 0.8 and 0.9 (NotFoundException relocated). So I would add that exception back, extend the new exception and throw it temporarily (also deprecate it).

dbu:
@sagikazarmark as long as we remove it for 1.0 i am +1 on that.

sagikazarmark:
I think we should remove it in 2.0 to make transition from 0.8 to 1.0 easier. Otherwise it's a BC break for consumer libraries.

dbu:
i would say caring a bit about BC for the 0.* versions to make upgrades
easier is fine, but 1.0 could be a cleanup. its not like this would be a
huge, hard to refactor problem. but if you insist, i can live with it.

sagikazarmark:
[...]

PR #69, opened 2016-07-03
Add back NotFoundException for BC

sagikazarmark:
As per the discussion here: https://github.com/php-http/discovery/issues/62#issuecomment-229590727

### DIFF
--- a/src/Exception/NotFoundException.php
+++ b/src/Exception/NotFoundException.php
@@ -5,7 +5,7 @@ namespace Http\Discovery\Exception;
 /**
  * Thrown when a discovery does not find any matches.
  *
- * @final
+ * @final do NOT extend this class, not final for BC reasons
  *
  * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
  */
--- a/src/Exception/NotFoundException.php
+++ b/src/Exception/NotFoundException.php
@@ -5,8 +5,10 @@ namespace Http\Discovery\Exception;
 /**
  * Thrown when a discovery does not find any matches.
  *
+ * @final
+ *
  * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
  */
-final class NotFoundException extends \RuntimeException
+/*final */class NotFoundException extends \RuntimeException
 {
 }
--- a/src/NotFoundException.php
+++ b/src/NotFoundException.php
@@ -2,8 +2,6 @@
 
 namespace Http\Discovery;
 
-@trigger_error('The '.__NAMESPACE__.'\NotFoundException class is deprecated since version 1.0 and will be removed in 2.0. Use Http\Discovery\Exception\NotFoundException instead.', E_USER_DEPRECATED);
-
 /**
  * Thrown when a discovery does not find any matches.
  *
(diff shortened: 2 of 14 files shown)


src/NotFoundException.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Http\Discovery;
+
+@trigger_error('The '.__NAMESPACE__.'\NotFoundException class is deprecated since version 1.0 and will be removed in 2.0. Use Http\Discovery\Exception\NotFoundException instead.', E_USER_DEPRECATED);
```
Nyholm: Do we need the __NAMESPACE__ constant here?

Nyholm:
👍 Good, I had just one minor question.

src/Exception/NotFoundException.php
```diff
@@ -5,8 +5,10 @@
 /**
  * Thrown when a discovery does not find any matches.
  *
+ * @final
+ *
  * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
  */
-final class NotFoundException extends \RuntimeException
+/*final */class NotFoundException extends \RuntimeException
```
dbu: maybe add "(do not extend this class, it is only not final for BC reasons)"

src/StreamFactoryDiscovery.php
```diff
@@ -3,7 +3,6 @@
 namespace Http\Discovery;
 
 use Http\Discovery\Exception\DiscoveryFailedException;
-use Http\Discovery\Exception\NotFoundException;
```
dbu: hm. should we have the phpdoc say it throws the new exception? people should start relying on the new exception, not the deprecated one.

src/NotFoundException.php
dbu: i think as long as we ourselves throw the old exception, we can't trigger a warning.

RE: namespace or not: does not matter imo. writing the actual namespace seems more straightforward though

sagikazarmark: I took this namespace thing from symfony. Since the namespace declaration is just above the warning, I think we are allowed to make it shorter in the code. The message itself is what's important after all.

But you are right, maybe we shouldn't throw this as long as we actually throw this exception.

src/Exception/NotFoundException.php
sagikazarmark: Isn't this what the `@final` phpdoc is for?

sagikazarmark: Isn't this what the `@final` phpdoc is for?

dbu: ah, makes sense. maybe add that comment after the `@final` then?i think explanations always are a good thing.

src/StreamFactoryDiscovery.php
sagikazarmark: done

src/Exception/NotFoundException.php
sagikazarmark: donre

dbu:
looks good to me. squash and merge?

sagikazarmark:
Yup, but rebase first.

PR #71, opened 2016-07-03
Bugfixes

Nyholm:
#Sorry for this PR. I've made some mistakes in previous PRs. 

Fix 1: 
We should see if the candidate got a condition when we fetch it from the cache. 

Fix 2: 
Improved the doc blocs

Fix 3: 
We should not require an object of type `DiscoveryStrategy` on `appendStratey` and `prependStrategy`. We should require a string FQCN.

### DIFF
--- a/src/ClassDiscovery.php
+++ b/src/ClassDiscovery.php
@@ -4,7 +4,6 @@ namespace Http\Discovery;
 
 use Http\Discovery\Exception\DiscoveryFailedException;
 use Http\Discovery\Exception\StrategyUnavailableException;
-use Http\Discovery\Strategy\DiscoveryStrategy;
 
 /**
  * Registry that based find results on class existence.
@@ -109,7 +108,7 @@ abstract class ClassDiscovery
     /**
      * Set new strategies and clear the cache.
      *
-     * @param DiscoveryStrategy[] $strategies
+     * @param array $strategies string array of FQN to a DiscoveryStrategy
      */
     public static function setStrategies(array $strategies)
     {
@@ -120,9 +119,9 @@ abstract class ClassDiscovery
     /**
      * Append a strategy at the end of the strategy queue.
      *
-     * @param DiscoveryStrategy $strategy
+     * @param string $strategy FQN to a DiscoveryStrategy
      */
-    public static function appendStrategy(DiscoveryStrategy $strategy)
+    public static function appendStrategy($strategy)
     {
         self::$strategies[] = $strategy;
         self::clearCache();
@@ -131,9 +130,9 @@ abstract class ClassDiscovery
     /**
      * Prepend a strategy at the beginning of the strategy queue.
      *
-     * @param DiscoveryStrategy $strategy
+     * @param string $strategy FQN to a DiscoveryStrategy
      */
-    public static function prependStrategy(DiscoveryStrategy $strategy)
+    public static function prependStrategy($strategy)
     {
         self::$strategies = array_unshift(self::$strategies, $strategy);
         self::clearCache();
--- a/src/ClassDiscovery.php
+++ b/src/ClassDiscovery.php
@@ -87,8 +87,10 @@ abstract class ClassDiscovery
         }
 
         $candidate = self::$cache[$type];
-        if (!self::evaluateCondition($candidate['condition'])) {
-            return;
+        if (isset($candidate['condition'])) {
+            if (!self::evaluateCondition($candidate['condition'])) {
+                return;
+            }
         }
 
         return $candidate['class'];
@@ -134,7 +136,7 @@ abstract class ClassDiscovery
      */
     public static function prependStrategy($strategy)
     {
-        self::$strategies = array_unshift(self::$strategies, $strategy);
+        array_unshift(self::$strategies, $strategy);
         self::clearCache();
     }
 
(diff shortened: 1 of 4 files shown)


src/ClassDiscovery.php
```diff
@@ -109,7 +110,7 @@ private static function storeInCache($type, $class)
     /**
      * Set new strategies and clear the cache.
      *
-     * @param DiscoveryStrategy[] $strategies
+     * @param array $strategies string array of FQN to a DiscoveryStrategy
```
sagikazarmark: FQCN you mean?

dbu: i'd prefer to write it out anyways :-)

`@param string[] $strategies Fully qualified class name to a DiscoveryStrategy

src/ClassDiscovery.php
```diff
@@ -120,9 +121,9 @@ public static function setStrategies(array $strategies)
     /**
      * Append a strategy at the end of the strategy queue.
      *
-     * @param DiscoveryStrategy $strategy
+     * @param string $strategy FQN to a DiscoveryStrategy
      */
-    public static function appendStrategy(DiscoveryStrategy $strategy)
+    public static function appendStrategy($strategy)
```
dbu: could you add a test? i think we should have a test that shows this was a problem.

Nyholm: You are correct. But Im not sure how to write a test for that. `self::$strategies` is private.

src/ClassDiscovery.php
```diff
@@ -131,11 +132,11 @@ public static function appendStrategy(DiscoveryStrategy $strategy)
     /**
      * Prepend a strategy at the beginning of the strategy queue.
      *
-     * @param DiscoveryStrategy $strategy
+     * @param string $strategy Fully qualified class name to a DiscoveryStrategy
      */
-    public static function prependStrategy(DiscoveryStrategy $strategy)
+    public static function prependStrategy($strategy)
```
dbu: maybe a full functional test with a prepend strategy that returns a mock?

src/ClassDiscovery.php
```diff
@@ -109,7 +110,7 @@ private static function storeInCache($type, $class)
     /**
      * Set new strategies and clear the cache.
      *
-     * @param DiscoveryStrategy[] $strategies
+     * @param array $strategies string array of fully qualified class name to a DiscoveryStrategy
      */
     public static function setStrategies(array $strategies)
```
dbu: a similar test could be done here, setting the strategies to only that test strategy that returns a mock client

src/ClassDiscovery.php
```diff
@@ -120,9 +121,9 @@ public static function setStrategies(array $strategies)
     /**
      * Append a strategy at the end of the strategy queue.
      *
-     * @param DiscoveryStrategy $strategy
+     * @param string $strategy Fully qualified class name to a DiscoveryStrategy
      */
-    public static function appendStrategy(DiscoveryStrategy $strategy)
+    public static function appendStrategy($strategy)
```
dbu: and to test this, first setStrategies to empty or something that never works, then append the one that returns a mock client.

src/ClassDiscovery.php
sagikazarmark: Add a resource ONLY in an appended strategy and see if you can resolve that resource.

src/ClassDiscovery.php
Nyholm: Please correct me if Im wrong. But the problem here is that `appendStrategy($strategy)` takes a FQCN not an object. I could add another `DiscoveryHelper` but it feels weird.. doesn't it?

Nyholm: Something like this is not possible. 

```php
    function it_prepends_strategies(DiscoveryStrategy $successfulStrategy, DiscoveryStrategy $addedStrategy) {
        $successfulStrategy->getCandidates('Foobar')->willReturn([['class'=>'Success']]);
        $addedStrategy->getCandidates('Foobar')->shouldBeCalled()->willReturn([['class'=>'Added']]);

        ClassDiscovery::setStrategies([$successfulStrategy]);
        ClassDiscovery::prependStrategy($addedStrategy);

        $this->find('Foobar')->shouldReturn('Added');
    }

    function it_appends_strategies(DiscoveryStrategy $failedStrategy, DiscoveryStrategy $addedStrategy) {
        $failedStrategy->getCandidates('Foobar')->shouldBeCalled()->willReturn([]);
        $addedStrategy->getCandidates('Foobar')->shouldBeCalled()->willReturn([['class'=>'Added']]);

        ClassDiscovery::setStrategies([$failedStrategy]);
        ClassDiscovery::appendStrategy($addedStrategy);

        $this->find('Foobar')->shouldReturn('Added');
    }
```

dbu: maybe like this:

```php
ClassDiscovery::setStrategies(['Http\Discovery\MockDiscovery']);
...
class MockDiscovery implements Discovery
{
    ...
}
class MockClient implements Client
{
    public function ... () {} // do nothing
}
```

but indeed with passing strings, you would need quite a few mock classes.

dbu:
LGTM!

Nyholm:
Thank you for merging.

PR #74, opened 2016-07-14
Allow for strategies to be objects or callables.

Nyholm:
This is just an idea. Im not sure if we want to do this or not but I would like to hear your opinions. 

Do we want to allow that strategies are objects or callables? This might be overkill when we allow candidates to be callables. 

Also, the DiscoveryStrategy interface says that `getCandidates` should be static. But calling it in a non static way is okey. https://3v4l.org/IbRNo

### DIFF
--- a/src/ClassDiscovery.php
+++ b/src/ClassDiscovery.php
@@ -5,6 +5,7 @@ namespace Http\Discovery;
 use Http\Discovery\Exception\ClassInstantiationFailedException;
 use Http\Discovery\Exception\DiscoveryFailedException;
 use Http\Discovery\Exception\StrategyUnavailableException;
+use Http\Discovery\Strategy\DiscoveryStrategy;
 
 /**
  * Registry that based find results on class existence.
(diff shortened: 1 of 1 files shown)


src/ClassDiscovery.php
```diff
@@ -51,7 +52,15 @@ protected static function findOneByType($type)
         $exceptions = [];
         foreach (self::$strategies as $strategy) {
             try {
-                $candidates = call_user_func($strategy.'::getCandidates', $type);
+                if (is_callable($strategy)) {
+                    $candidates = $strategy($type);
+                } elseif ($strategy instanceof DiscoveryStrategy) {
+                    $candidates = $strategy->getCandidates($type);
+                } elseif (is_string($strategy)) {
```
sagikazarmark: Shouldn't we check an instance of DiscoveryStrategy too?

dbu:
i personally think we should not invest much into discovery and promote
dependency injection with explicit instantiation instead. otherwise this
ends up being a global service container thing.

that said, if there is a use case for this it looks like the right way.

Nyholm:
The reason why I've been back and forth with this is because Im not 100% happy with this: https://github.com/php-http/HttplugBundle/pull/85#issuecomment-232411127

But Im not sure this PR will create a possibility to make https://github.com/php-http/HttplugBundle/pull/85 better. I know this is very vague..

Nyholm:
No, changed my mind with this for good. 

Since we allow closures as candidates once could easy create something like this:
```php
class Strategy implements DiscoveryStrategy {

  private $object; 
  public function __construct($object) { 
    $this->object = $object;
  }
  static function getCandidates($type) {
     $obj = $this-object;
     return [[
          'class'=>function() use ($obj) {
              return $obj->getClass($type);
          }, 
          'condition'=>function() use ($obj) {
              return $obj->evaluateCondition($type);
          }, 
       ]];
  }
}

PR #75, opened 2016-07-15
Prepare release 1.0

Nyholm:
Just a clarification in the changelog.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
     ],
     "extra": {
         "branch-alias": {
-            "dev-master": "0.10-dev"
+            "dev-master": "1.0-dev"
         }
     },
     "config": {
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -43,7 +43,7 @@
     ],
     "extra": {
         "branch-alias": {
-            "dev-master": "0.10-dev"
+            "dev-master": "1.1-dev"
```
dbu: we did not tag 1.0.0 - i would only change this once we add a new feature to discovery.

Nyholm: Okey, I thought you changed this before you made the tag. Sorry about that.

sagikazarmark: AFAIK this should be changed to 1.0-dev as we did in other packages

dbu: ups, yes. i agree. sorry, somehow missed that its currently 0.10 and not 1.0

Issue #48, opened 2016-01-15
Remove puli.phar


sagikazarmark:
Once puli/issues#165 is done

sagikazarmark:
Now that Puli is optional, should we remove the binary from the package? Would it be a BC break if we did it later?

Nyholm:
I think we should remove it. 
It would never be a BC break beacuse the discovery strategies are not part of the BC promise. IE we could remove the PuliStrategy without breaking BC....

This is a bit weird when I think about it... I'll document it. But yes, Puli.phar should be removed before 1.0.

dbu:
fixed in #76

PR #76, opened 2016-07-16
Improve package

sagikazarmark:


sagikazarmark:
Okay, so this PR improves some package files, fixes styleci config AND removes the puli binary as discussed in #48. Please review.

CONTRIBUTING
```diff
@@ -1 +0,0 @@
-Please see http://docs.php-http.org/en/latest/development/contributing.html
```
Nyholm: This means that GitHub does not show this "warning" anymore. 
screen shot 2016-07-16 at 12 02 30

composer.json
```diff
@@ -11,14 +11,14 @@
         }
     ],
     "require": {
-        "php": "^5.5|^7.0"
+        "php": "^5.5 || ^7.0"
     },
     "require-dev": {
         "php-http/httplug": "^1.0",
         "php-http/message-factory": "^1.0",
         "puli/composer-plugin": "1.0.0-beta9",
```
Nyholm: Let's use the latest beta release. Beta10

CONTRIBUTING
sagikazarmark: It is moved to `.github`

Nyholm: Sorry, I did not know you could put it there. All good then!

composer.json
sagikazarmark: Do we still support beta9? If so, I would rather use `^1.0.0-beta9`.

Nyholm: If our strategy works with beta9 sure. We support it. (I think it does)

Im not sure `^1.0.0-beta9` is the same as `1.0.0-beta9 || 1.0.0-beta10`. And if it is, we should not use it. Because we can guarantee it works with `1.0.0-beta11`

sagikazarmark: Let's use `1.0.0-beta9 || 1.0.0-beta10` then.

sagikazarmark:
Okay, need someone else too to review this as the final PR before tagging a stable. /cc @dbu

composer.json
```diff
@@ -11,14 +11,14 @@
...
     ],
     "require": {
-        "php": "^5.5|^7.0"
+        "php": "^5.5 || ^7.0"
     },
     "require-dev": {
         "php-http/httplug": "^1.0",
         "php-http/message-factory": "^1.0",
-        "puli/composer-plugin": "1.0.0-beta9",
+        "puli/composer-plugin": "1.0.0-beta9 || 1.0.0-beta10",
```
dbu: this is dev dependency anyways, do we need to allow multiple versions?

dbu: i merged anyways, as this is really not important. if you want to change it, plesae do. otherwise i think it does no harm.

sagikazarmark: Indeed. Should we add conflict field saying these are the only supported versions?

Nyholm: I do not think so. It is in the docs. (My BC pr)

If we want to do this properly we should config Travis to run tests with different versions of puli.

dbu: we could mention it in the suggest section. i would not add conflicts.

sagikazarmark: Ok, will do that.

PR #77, opened 2016-07-16
Added a BC note

Nyholm:
It should be more clear what our BC is. With this BC promise we can. 

* Change the order of our strategies and add strategies in any order
* Change the order/priority of classes in CommonClassesStrategy

The consequence is that you *can never* be sure **what** client that is going to be found but you *can* be sure that updating a minor version will not result in that a client is not found any more.  

Related to #48

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,13 @@ Please see the [official documentation](http://php-http.readthedocs.org/en/lates
 $ composer test
 ```
 
+## Backwards compatibility promise
+
+The backwards compatibility promise does not include classes marked with the `@internal` annotation. Nor does it
+include the default order of which the strategies are executed. However we do promise that we do not remove any strategies 
+and we will also not remove any classes in the `CommonClassesStrategy`. We do also promise to support the following Puli versions: 
+
+* 1.0.0-beta10
 
 ## Contributing
 


README.md
```diff
@@ -37,6 +37,13 @@ Please see the [official documentation](http://php-http.readthedocs.org/en/lates
 $ composer test
 ```
 
+## Backwards compatibility promise
+
+The backwards compatibility promise does not include classes marked with the `@internal` annotation. Nor does it
+include the default order of which the strategies are executed. However we do promise that we do not remove any strategies 
+and we will also not remove any classes in the `CommonClassesStrategy`. We do also promise to support the following Puli versions:
```
sagikazarmark: I would rather say "we promise to support/not remove the following classes: [list]".

sagikazarmark:
Shouldn't we rather place this in the documentation?

README.md
Nyholm: I don't want to make a list because it will be harder to maintain (especially if we move this to the documentation repo). But the formulation could be improved.

Nyholm:
Maybe it should be placed in the documentation repo. I just figured, if you are about to update versions you are most probable to go to packagist, github and then to our documentation page. 

In one way it makes sense to have the  BC promise close to the changelog and upgrade docs. But it do also make a lot of sense to have the BC promise next to all our other documentation. 

Maybe we should create a full page in the docs about our general BC promise and all our special ones (like this and possible also for HTTPlugBundle).

sagikazarmark:
Sounds like a good idea to me.

Tobias Nyholm <notifications@github.com> ezt írta (2016. július 16.,
szombat):

PR #78, opened 2016-07-18
Test with multiple puli versions

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

#### What's in this PR?

Tests the package against multiple Puli versions.

#### Why?

Ensure that all beta versions are working.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,9 @@ php:
 env:
     global:
         - TEST_COMMAND="composer test"
+    matrix:
+        - PULI_VERSION=1.0.0-beta9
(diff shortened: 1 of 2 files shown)


Nyholm:
I do not think we need to test each PHP version with each Puli version. We have to assume that Puli is tested properly with different PHP versions already.

sagikazarmark:
Yeah, and the test matrix became insanely huge, so let's just assume it's working. :smile:

Nyholm:
Wait.. 
Let me try someting

PR #85, opened 2016-11-25
Using 'self' instead of 'static'

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | n/a
| Documentation   | n/a
| License         | MIT

#### Why

Scrutinizer suggest to change `self` to `static`. Since the `CommonClassesStrategy` is make no sense to have `static`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## Unreleased
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


Nyholm:
Thank you for merging

PR #90, opened 2017-01-27
Add Mock Client strategy.

samrap:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    |yes|
| BC breaks?      | no|
| Deprecations?   | no|
| Related tickets | mentioned in #89 
| Documentation   | https://github.com/php-http/documentation/pull/173
| License         | MIT

#### What's in this PR?

Please see #89 for a full description of this PR.

Essentially, we need a way to test DI within our tests which in many cases cannot be fully implemented without having some way to discover the Mock Client. This new class allows us to add the `MockClientStrategy` to the `HttpClientDiscovery`'s strategies as shown in @Nyholm 's example in the above mentioned PR.

#### Example Usage

``` php
use ThirdPartyService

class ThirdPartyServiceTest extends PHPUnit_Framework_TestCase
{
    public function setUpBeforeClass()
    {
        HttpClientDiscovery::prependStrategy(MockClientStrategy::class);
    }

    public function setUp()
    {
        $this->service = new ThirdPartyService;
    }
}
```

#### Checklist

- [x] Documentation pull request created (if not simply a bugfix)

### DIFF
--- a/src/Strategy/MockClientStrategy.php
+++ b/src/Strategy/MockClientStrategy.php
@@ -7,9 +7,6 @@ use Http\Mock\Client as Mock;
 /**
  * Find the Mock client.
  *
- * @internal
- * @final
- *
  * @author Sam Rapaport <me@samrapdev.com>
  */
 final class MockClientStrategy implements DiscoveryStrategy
--- a/src/Strategy/MockClientStrategy.php
+++ b/src/Strategy/MockClientStrategy.php
@@ -2,6 +2,7 @@
 
 namespace Http\Discovery\Strategy;
 
+use Http\Client\HttpClient;
 use Http\Mock\Client as Mock;
 
 /**
@@ -16,7 +17,7 @@ final class MockClientStrategy implements DiscoveryStrategy
      */
     public static function getCandidates($type)
     {
-        return ($type === 'Http\Client\HttpClient')
+        return ($type === HttpClient::class)
             ? ['class' => Mock::class, 'condition' => Mock::class]
             : [];
     }
--- a/src/Strategy/MockClientStrategy.php
+++ b/src/Strategy/MockClientStrategy.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace Http\Discovery\Strategy;
+
+use Http\Mock\Client as Mock;
+
+/**
+ * Find the Mock client.
+ *
+ * @internal
+ * @final
+ *
+ * @author Sam Rapaport <me@samrapdev.com>
+ */
+final class MockClientStrategy implements DiscoveryStrategy
+{
+    /**
+     * {@inheritdoc}
+     */
+    public static function getCandidates($type)
+    {
+        return ($type === 'Http\Client\HttpClient')
+            ? ['class' => Mock::class, 'condition' => Mock::class]
+            : [];
+    }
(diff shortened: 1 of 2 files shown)


src/Strategy/MockClientStrategy.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Http\Discovery\Strategy;
+
+use Http\Mock\Client as Mock;
+
+/**
+ * Find the Mock client.
+ *
+ * @internal
+ * @final
```
Nyholm: No need for final annotation

src/Strategy/MockClientStrategy.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Http\Discovery\Strategy;
+
+use Http\Mock\Client as Mock;
+
+/**
+ * Find the Mock client.
+ *
+ * @internal
```
Nyholm: Is **this strategy** really internal? The others are, but it is meant that this class should be used by others, right?

src/Strategy/MockClientStrategy.php
```diff
@@ -0,0 +1,26 @@
...
+ * @author Sam Rapaport <me@samrapdev.com>
+ */
+final class MockClientStrategy implements DiscoveryStrategy
+{
+    /**
+     * {@inheritdoc}
+     */
+    public static function getCandidates($type)
+    {
+        return ($type === 'Http\Client\HttpClient')
```
sagikazarmark: You could use the ::class constant here

sagikazarmark:
What do you think about https://github.com/php-http/discovery/pull/89#issuecomment-275753774 ?

src/Strategy/MockClientStrategy.php
samrap: 👍

src/Strategy/MockClientStrategy.php
samrap: Good call, changing that now.

samrap:
@sagikazarmark I prefer sticking with a manual setup in the test class i.e:

```php
public function setUp()
{
    HttpClientDiscovery::prependStrategy(MockClientStrategy::class);
}
```

Adding a trait or method(s) to the Mock Client to add itself to the Discovery means the strategy now depends on the Discovery which seems like an unnecessary dependency to me. Unless I'm understanding this wrong.

sagikazarmark:
> Adding a trait or method(s) to the Mock Client to add itself to the

I rather meant a trait or base class for PHPUnit test cases, but I realized that MockClient is not limited to be used in PHPUnit, so that's probably not really good idea.

Nyholm:
I do not like Discovery::enableMockClient. No discovery strategy should be treated in special manners.

samrap:
@Nyholm good point. So I think we should keep this as is, and I will work on getting those docs completed.

samrap:
Just FYI still active on this, been busy at work but will get these docs by end of week.

samrap:
See https://github.com/php-http/documentation/pull/173

Nyholm:
Thank you for this PR.

samrap:
Thank you guys!

PR #92, opened 2017-02-28
Fix mock client strategy

captn3m0:
| Q               | A|
| --------------- | ---|
| Bug fix?        | yes
| New feature?    | no|
| BC breaks?      | no|
| Deprecations?   | no|
| Related tickets | -| 
| Documentation   | -|
| License         | MIT

#### What's in this PR?

- The Strategy returned a invalid array earlier.

>The return value is always an array with zero or more elements.
>Each  element is an array with two keys
>['class' => string, 'condition' => mixed]

<detail><summary>The older value would lead to `Illegal string offset 'class'` error:</summary>

```
Illegal string offset 'class'

/vendor/php-http/discovery/src/ClassDiscovery.php:70
/vendor/php-http/discovery/src/HttpClientDiscovery.php:25
/src/php/Client.php:22
/tests/ClientTest.php:18
```
</detail>

#### Why?

Mock Client Stategy Discovery is broken without this fix.

#### Checklist

#### To Do

I'm not used to phpspec, so I'm not sure if the spec suffices here, or what can be done to improve it. I attempted using HttpClientDiscovery with the strategy set to mock, but that required adding too many dependencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 
 ## Unreleased
 
+- MockClientStategy class fixed
+
 ## 1.2.0 - 2017-02-12
 
 ### Added
--- a/spec/Strategy/MockClientStrategySpec.php
+++ b/spec/Strategy/MockClientStrategySpec.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace spec\Http\Discovery\Strategy;
+
+use Http\Client\HttpClient;
+use Http\Discovery\ClassDiscovery;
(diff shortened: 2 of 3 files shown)


sagikazarmark:
LGTM

dbu:
thank you!

captn3m0:
Any chance this can get a minor version bump?

Nyholm:
No, but you should get a patch release. I will do that tonight unless someone else beat me to it.

PR #100, opened 2017-08-19
Use ::class constants

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

#### What's in this PR?

Using ::class constants for all class expressions

#### Why?

Im working with [php-scoper](https://github.com/humbug/php-scoper). It rewrites all namespaces with a prefix. Since we are using a simple string (not a class) php-scoper has no change of rewriting this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## Unreleased
+
+### Changed
+
+- Use ::class constant instead of writing the class string. 
+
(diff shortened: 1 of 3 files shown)


purpleweb:
:+1:

Nyholm:
Thank you for merging

PR #101, opened 2017-09-13
Update composer add support of PHP 5.6

purpleweb:
This PR just add support of PHP 5.6 in composer.
Thanks for your work !

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         }
     ],
     "require": {
-        "php": "^5.5 || ^7.0"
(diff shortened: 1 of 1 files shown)


fbourigault:
This is not needed as `^5.5`means `>=5.5.0 <6.0.0` See https://getcomposer.org/doc/articles/versions.md#caret-version-range-.

purpleweb:
Ok sorry you're right. I had an installation problem, but that was something else !
Thank you !

PR #108, opened 2018-08-01
Updated strategy for Nyholm/psr7

Nyholm:
This is needed for 1.0

See https://github.com/Nyholm/psr7/pull/62


sagikazarmark:
Should we add a conflict config to composer, so that the next version of discovery can't be used with nyholm/psr <1.0?

Nyholm:
True, we could do that. Currently we support all versions of nyholm/psr7. But your idea sounds better.

Nyholm:
Yes, this is better. Thank you

Nyholm:
Version 1.0 is released now. 

@sagikazarmark are you happy with this PR?

Nyholm:
Could I get a review please

PR #109, opened 2018-08-01
Support for Buzz 1.0

Nyholm:
This in WIP because we have not tagged 1.0 of Buzz yet.

### DIFF
--- a/src/Strategy/CommonClassesStrategy.php
+++ b/src/Strategy/CommonClassesStrategy.php
@@ -78,7 +78,7 @@ final class CommonClassesStrategy implements DiscoveryStrategy
             ['class' => Zend::class, 'condition' => Zend::class],
             ['class' => Artax::class, 'condition' => Artax::class],
             [
-                'class' => ['CommonClassesStrategy::buzzInstanciate',
+                'class' => ['CommonClassesStrategy::buzzInstanciate'],
                 'condition' => [\Buzz\Client\Curl::class, \Buzz\Message\ResponseBuilder::class]
             ],
         ],
--- a/src/Strategy/CommonClassesStrategy.php
+++ b/src/Strategy/CommonClassesStrategy.php
@@ -78,9 +78,7 @@ final class CommonClassesStrategy implements DiscoveryStrategy
             ['class' => Zend::class, 'condition' => Zend::class],
             ['class' => Artax::class, 'condition' => Artax::class],
             [
-                'class' => function() {
-                    return new \Buzz\Client\Curl([], MessageFactoryDiscovery::find());
-                },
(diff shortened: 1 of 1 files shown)


Nyholm:
This is not ready for a review.

src/Strategy/CommonClassesStrategy.php
```diff
@@ -77,6 +78,10 @@
             ['class' => Cake::class, 'condition' => Cake::class],
             ['class' => Zend::class, 'condition' => Zend::class],
             ['class' => Artax::class, 'condition' => Artax::class],
+            [
+                'class' => [self::class, 'buzzInstantiate'],
+                'condition' => [\Buzz\Client\FileGetContents::class, \Buzz\Message\ResponseBuilder::class],
```
dbu: are those in separate packages? or is it a bit of a hack to see if a new enough version of buzz is available?

src/Strategy/CommonClassesStrategy.php
```diff
@@ -91,4 +96,9 @@ public static function getCandidates($type)
 
         return [];
     }
+
+    public static function buzzInstantiate()
+    {
+        return new \Buzz\Client\FileGetContents(MessageFactoryDiscovery::find());
```
dbu: can buzz work with any message factory? there is the super edge case where somebody is using discovery to find a client but then instantiates their own message implementations so we can't find a message factory.

as buzz is the last client in order, that seems irrelevant to me, however. the hyper-correct thing would be to add a condition callback function above that makes sure message factory discovery can find a message factory...

src/Strategy/CommonClassesStrategy.php
Nyholm: It is a version hack =)

src/Strategy/CommonClassesStrategy.php
Nyholm: Yes, Buzz supports PSR17 and Httplug message factories. 

Hm, that is true. 

Btw, buzz does not have "their own factories". So it would be quite normal that someone installed buzz and not having any message factories installed.

Nyholm: > the hyper-correct thing would be to add a condition callback function above that makes sure message factory discovery can find a message factory...

I agree with you. Then I spend 20 minutes on implementation and I realized that we dont do this for any other client. See: 
https://github.com/php-http/curl-client/blob/master/src/Client.php#L79

dbu: :man_facepalming: sorry about that. yeah lets not, then.

Nyholm:
Thank you for merging

Nyholm:
Yes, that package should be deprecated.

PR #110, opened 2018-08-17
Introdce a safe_class_exists

### LINKED ISSUE #107: Composer dependency missing?
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | no
| Version      | master

Just a quick question: In `Http\Discovery\Strategy\CommonClassesStrategy` you use classes i.e. `Nyholm\Psr7\Request` and this dependency is not defined in the `composer.json`. How does it work? This package composer.json states:
```
"require": {
        "php": "^5.5 || ^7.0"
    }
```

The Nyholm composer.json (https://github.com/Nyholm/psr7/blob/master/composer.json) states:
```
"require": {
        "php": "^7.1",
```

How can I use this package without conflicts and how is it supposed to be used when the dependencies are not explicitly declared? Thanks a lot guys.

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #107
| Documentation   | 
| License         | MIT

#### What's in this PR?

A workaround to make it work with Magento. 

This solution to introduce a functions.php is inspired by Guzzle/PSR7. Now when Im done with the implementation I think it may be overkill.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,8 @@
     "autoload": {
         "psr-4": {
             "Http\\Discovery\\": "src/"
-        }
+        },
+        "files": ["src/functions_include.php"]
     },
     "autoload-dev": {
         "psr-4": {
--- a/src/functions_include.php
+++ b/src/functions_include.php
@@ -2,5 +2,5 @@
 
 // Don't redefine the functions if included multiple times.
 if (!function_exists('Http\Discovery\safe_class_exists')) {
-    require __DIR__ . '/functions.php';
+    require __DIR__.'/functions.php';
 }
--- a/src/functions_include.php
+++ b/src/functions_include.php
@@ -1,0 +1,6 @@
+<?php
+
+// Don't redefine the functions if included multiple times.
+if (!function_exists('Http\Discovery\safe_class_exists')) {
+    require __DIR__ . '/functions.php';
+}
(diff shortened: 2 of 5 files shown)


src/functions_include.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+// Don't redefine the functions if included multiple times.
+if (!function_exists('Http\Discovery\safe_class_exists')) {
```
dbu: why not require_once?

composer.json
```diff
@@ -27,7 +27,8 @@
     "autoload": {
         "psr-4": {
             "Http\\Discovery\\": "src/"
-        }
+        },
+        "files": ["src/functions_include.php"]
```
dbu: is there no autoloading for functions?

src/functions_include.php
Nyholm: I *think* this allows you to override... Im not sure. I just did it the "guzzle" way.

composer.json
Nyholm: I think only classes are autoloaded.

dbu: right, otherwise we would need a naming scheme for function to filename...

src/functions_include.php
dbu: yeah that allows to override by hijacking the namespace. sounds like either horrible situations or really bad design but i guess it does not hurt too much.

also will be quite some issue with having things loaded in the right order. but if its helpful to people to have it, i guess the overhad is insignificant so lets do it this way.

Nyholm:
Can you also check the alternative implementation for this feature: #111

dbu:
ah, now i know why i was confused. i prefer #111 over the function, it seems less complicated infrastructure to me.

Nyholm:
Yeah, I agree with you.

PR #121, opened 2018-12-29
Support for discovering psr-18 clients

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

Allow us to discover psr-18 clients

### DIFF
--- a/src/Psr18ClientDiscovery.php
+++ b/src/Psr18ClientDiscovery.php
@@ -6,7 +6,7 @@ use Http\Discovery\Exception\DiscoveryFailedException;
 use Psr\Http\Client\ClientInterface;
 
 /**
- * Finds a PSR18 HTTP Client.
+ * Finds a PSR-18 HTTP Client.
  *
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
  */
--- a/src/Psr18ClientDiscovery.php
+++ b/src/Psr18ClientDiscovery.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Http\Discovery;
+
+use Http\Discovery\Exception\DiscoveryFailedException;
+use Psr\Http\Client\ClientInterface;
+
+/**
+ * Finds a PSR18 HTTP Client.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+final class Psr18ClientDiscovery extends ClassDiscovery
+{
+    /**
+     * Finds a PSR-18 HTTP Client.
+     *
+     * @return ClientInterface
+     *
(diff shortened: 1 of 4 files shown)


src/Strategy/CommonClassesStrategy.php
```diff
@@ -82,13 +83,25 @@
...
+        Psr18Client::class => [],
     ];
 
     /**
      * {@inheritdoc}
      */
     public static function getCandidates($type)
     {
+        if (Psr18Client::class === $type) {
+            $candidates = self::$classes[$type];
```
dbu: i would prefer to say `$candiates = self::$classes[PSR18Client::class];` to be more explicit as to what we do here.

src/Strategy/CommonClassesStrategy.php
```diff
@@ -82,13 +83,25 @@
...
         ],
+        Psr18Client::class => [],
     ];
 
     /**
      * {@inheritdoc}
      */
     public static function getCandidates($type)
     {
+        if (Psr18Client::class === $type) {
```
dbu: add a comment that we merge the httplug clients that implement psr18 too?

src/Strategy/CommonClassesStrategy.php
Nyholm: Good!

src/Strategy/CommonClassesStrategy.php
Nyholm: Sure

dbu:
i think i like the grouping by psr, i would not change what we did for psr 17.

Psr18ClientDiscovery::find would be nice i think. (its a client, not a factory)

src/Psr18ClientDiscovery.php
```diff
@@ -0,0 +1,36 @@
...
+     * @return ClientInterface
+     *
+     * @throws Exception\NotFoundException
+     */
+    public static function find()
+    {
+        try {
+            $client = static::findOneByType(ClientInterface::class);
+        } catch (DiscoveryFailedException $e) {
+            throw new \Http\Discovery\Exception\NotFoundException(
```
Nyholm: Since `Http\Discovery` contains a class called `NotFoundException` I cannot use a `use` statement. And using an alias is more confusing then just be explicit.

Nyholm:
Thank you for the reviews

src/Psr18ClientDiscovery.php
dbu: fair enough :+1:

PR #122, opened 2018-12-29
Added a discovery of classes considered as deprecated

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   |
| License         | MIT

According to our [BC promise](https://php-http.readthedocs.io/en/latest/httplug/backwards-compatibility.html#discovery) we can never remove previously discovered classes. 

In our current master we have removed support for nyholm/psr7:0.3.0. Im not sure nobody will notice that we break BC here. But I think we should care.

That is why I added a new strategy, just like `CommonClassesStrategy` but for "deprecated" or removed classes.

### DIFF
--- a/src/ClassDiscovery.php
+++ b/src/ClassDiscovery.php
@@ -24,6 +24,7 @@ abstract class ClassDiscovery
     private static $strategies = [
         Strategy\PuliBetaStrategy::class,
         Strategy\CommonClassesStrategy::class,
+        Strategy\DeprecatedClassesStrategy::class,
     ];
 
     /**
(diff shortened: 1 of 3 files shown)


Nyholm:
Oh, never mind. I just saw this commit: https://github.com/php-http/discovery/commit/a107c86fcf986e7dfcd3b347c919c4264dfe4af3

We say that we conflict with nyholm/psr7<1.0

dbu:
i think that conflict is good enough as BC. pre 1.0 versions are a special thing anyways...

PR #123, opened 2018-12-29
Removed code from nyholm/psr7 0.3.0

Nyholm:


Nyholm:
Yes, it has been released. But it is fine because we define a conflict. https://github.com/php-http/discovery/blob/master/composer.json#L47

I made a mistake in #108

Im not sure why this line was added there

dbu:
ah cool, the conflict declaration makes this safe.

Nyholm:
Thank you

PR #126, opened 2018-12-30
Removed PHP7 code

Nyholm:
Oops, I made a mistake. 

This happened because we never instantiate this class.


dbu:
lets tag 1.5.1 on this and then remove the branch again, i'd say.

Nyholm:
Could you really tag a commit and then remove the branch?

Hm. I’ll try

dbu:
yep you can. the tag is on the specific commit, not on a branch.

PR #129, opened 2019-01-02
Update Psr17FactoryDiscovery.php

gmponos:
Exception must be returned check CI here https://github.com/florianv/exchanger/pull/71


gmponos:
> ouch. are we missing some tests?

I guess.. I had intention of adding them once I find time...

dbu:
glad if you can. testing the "nothing found" is a bit tricky as we would need a separate build run. maybe we can overwrite what classes to try to discover for a test that finds nothing?

gmponos:
> as we would need a separate build run.

or use reflection on a test and handle by removing the classes from the property?

dbu:
that sounds legit to me too.

PR #130, opened 2019-01-02
Add PSR-17 to strategies

gmponos:
Shouldn't this be in here?

Found from this PR https://github.com/florianv/exchanger/pull/71 that I used the PSR-17 discovery


gmponos:
Also in the linked PR it does not find any PSR-18 client through discovery.. or I am doing something wrong?

Nyholm:
You are correct. Thanks!

If you have installed guzzle6-adapter or buzz you should be able to find them using dev-master of discovery.

Issue #138, opened 2019-03-01
Support the Symfony PSR-18 http client ?


Taluu:
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes
| Version      | master ?

#### Actual Behavior
The discovery doesn't discover the symfony http client, especially its psr-18 bridge. Would be great to support that IMO.

#### Expected Behavior
Being able to discover the PSR-18 adapter of the brand new Symfony HttpClient component.

I'll try a PR once the component is merged and released (or maybe before ?)

dbu:
sure, go ahead. we will not merge anything before the symfony component is released, but trying it out now is a good idea, to discover potential problems before it is released.

Taluu:
As this was implemented, I guess this can be now closed. Unless you'd like to keep this open until there's a release ? Feel free to reopen if that's the case.

PR #139, opened 2019-06-19
Add support for symfony/http-client

nicolas-grekas:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #138
| Documentation   | n/a
| License         | MIT

#### Checklist

- [x] Documentation pull request not needed

### DIFF
--- a/src/Strategy/CommonClassesStrategy.php
+++ b/src/Strategy/CommonClassesStrategy.php
@@ -5,7 +5,9 @@ namespace Http\Discovery\Strategy;
 use GuzzleHttp\Psr7\Request as GuzzleRequest;
 use Http\Client\HttpAsyncClient;
 use Http\Client\HttpClient;
+use Http\Discovery\ClassDiscovery;
 use Http\Discovery\MessageFactoryDiscovery;
+use Http\Discovery\Psr17FactoryDiscovery;
 use Http\Message\MessageFactory;
 use Http\Message\MessageFactory\GuzzleMessageFactory;
 use Http\Message\StreamFactory;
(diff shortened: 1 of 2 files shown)


src/Strategy/CommonClassesStrategy.php
```diff
@@ -104,7 +116,7 @@ public static function getCandidates($type)
                 }
             }
 
-            return $candidates;
+            return array_merge($candidates, self::$classes[Psr18Client::class]);
```
nicolas-grekas: this ensure that the order of discovery is the same for httplug and psr18 clients

dbu: with autodiscovery, it could be kind of tricky to change the ordering - people might rely on getting a specific client and this change could trip them up. if they refactor to psr-18 and then get a different client, i think that is less surprising than just updating discovery.

@Nyholm i think you wrote this originally. was the intention of the ordering to prefer the psr-18 clients when psr-18 is requested?

i would tend to revert this change. but can be convinced otherwise ;-)

nicolas-grekas: Since only buzz is listed explicitly in the psr18 list, this means ppl that have both buzz and something else will get the something else instead. People that don't have buzz won't see a difference.
But having Symfony before Guzzle when PSR18 is loaded, and Guzzle before Symfony when httplug is is more weird.

dbu: right. plus when we add to the PSR18 list, we might prepend another client (like the symfony client, which has some chance of being installed unwittingly when upgrading dependencies...). well anyways, discovery is the yolo approach :-)

waiting for @Nyholm to give his opinion too.

nicolas-grekas: reverted, I'll let you care in another PR if you do :)

PR #141, opened 2019-06-24
Moved Symfony before Buzz

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

I moved Symfony before Buzz since it is fairly more used and maintained. Im not sure if this list should be ordered by popularity, performance, most likeable or random. But this *feels* better.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ matrix:
           env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" COVERAGE=true TEST_COMMAND="composer test" PULI_VERSION=1.0.0-beta9
         - name: PHPUnit tests
           php: 7.3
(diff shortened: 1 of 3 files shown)


Nyholm:
Should we update order on the HTTPlug adapters too? @dbu @sagikazarmark?

sagikazarmark:
I'm okay if we suggest a better maintained client. It only matters anyway if there are multiple ones installed.

For the user it shouldn't really matter, they should rely on the interface. The only thing that would impact them is a client not adhering the standard or performance differences. Again, I think we can make the better choice for the user in those cases as well.

Nyholm:
Okey. Thank you for the input. I've updated the PR

Nyholm:
Thank you for the reviews

PR #145, opened 2019-09-23
Support PSR-18 interface as candidate by inheritance

### LINKED ISSUE #144: PSR-18 client discovery will not find the Mock client
| Q            | A
| ------------ | ---
| Bug?         |yes
| New Feature? | no

#### Actual Behavior

A DiscoveryFailedException is thrown.

#### Expected Behavior

An instance of the mock client.

#### Steps to Reproduce

Using the Psr18ClientDiscovery will result in a DiscoveryFailedException (MockClientStrategy is prepended) since the current MockClientStrategy have no candidate for the PSR ClientInterface.

#### Possible Solutions

Will open a PR soon.

pascalwoerde:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #144
| Documentation   | -
| License         | MIT

#### What's in this PR?

Support the PSR-18 interface as candidate by inheritance instead of two fixed class names.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ matrix:
           php: 7.1
           # Disable code coverage until https://github.com/leanphp/phpspec-code-coverage/pull/38 is released
           # DEPENDENCIES="leanphp/phpspec-code-coverage:^4.2" TEST_COMMAND="composer test-ci"
-          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" COVERAGE=true TEST_COMMAND="composer test" DEPENDENCIES="php-http/httplug:^2.0" PULI_VERSION=1.0.0-beta9
+          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" COVERAGE=true TEST_COMMAND="composer test" PULI_VERSION=1.0.0-beta9
         - name: PHPUnit tests
           php: 7.3
           env: TEST_COMMAND="./vendor/bin/phpunit" DEPENDENCIES="phpunit/phpunit:^7.5 nyholm/psr7:^1.0 kriswallsmith/buzz:^1.0 php-http/curl-client:^1.0 php-http/message"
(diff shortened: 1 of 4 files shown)


pascalwoerde:
> i am quite confused why the test fails when running with coverage, but succeeds in the other builds. @php-http/owners any ideas why this would happen, do you see what is wrong?

Ah, this seems to the `--prefer-lowest` option in `composer update` before running phpspec. Shall I rewrite the specification to unit test?

dbu:
ah, i missed that, good find! if its about the version of phpspec then you can also change the minimum version in require-dev.

.travis.yml
```diff
@@ -27,7 +27,7 @@ matrix:
           php: 7.1
           # Disable code coverage until https://github.com/leanphp/phpspec-code-coverage/pull/38 is released
           # DEPENDENCIES="leanphp/phpspec-code-coverage:^4.2" TEST_COMMAND="composer test-ci"
-          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" COVERAGE=true TEST_COMMAND="composer test" PULI_VERSION=1.0.0-beta9
+          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" COVERAGE=true TEST_COMMAND="composer test" DEPENDENCIES="php-http/httplug:^2.0" PULI_VERSION=1.0.0-beta9
```
dbu: this means we no longer test with httplug 1 at all, as all the other builds install newest.

spec/Strategy/MockClientStrategySpec.php
```diff
@@ -24,6 +25,13 @@ function it_should_return_the_mock_client(DiscoveryStrategy $strategy)
         $candidates->shouldHaveCount(1);
     }
 
+    function it_should_return_the_mock_client_for_psr_interface(DiscoveryStrategy $strategy)
+    {
+        $candidates = $this->getCandidates(ClientInterface::class);
```
dbu: could we do a class_exists here and pick the old or the new client interface to also be able to test this with httplug 1?

pascalwoerde: The specification is rewritten with `class_implements`, so that we can cover all the implemented interfaces for now and in the future. `class_exists` will not do the job for us, since this specification will depend on the version of our HttpClient interface from HTTPlug.

PR #153, opened 2019-12-20
Test on PHP 7.4

GrahamCampbell:
Run tests with PHP 7.4


dbu:
seems the version constraint for phpspec is preventing installation. can you please try allowing a phpspec version that can install with php 7.4 and then we see if other things break because of that?

PR #154, opened 2019-12-26
Allow a condition to specify an interface and not just classes.

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

We previously introduced "conditions" that specify interfaces. But we never made sure they could be evaluated to true.


Nyholm:
Thank you

PR #157, opened 2019-12-27
Added small script to test installation on with different dependencies

### LINKED ISSUE #156: [Symfony 4.4] No PSR-17 response factory found
It seems https://github.com/php-http/discovery/issues/149 is still not fixed. The fix that was proposed in https://github.com/php-http/discovery/pull/150 was reverted in https://github.com/php-http/discovery/pull/151

> No PSR-17 response factory found. Install a package from this list: https://packagist.org/providers/psr/http-factory-implementation

>  Could not find resource using any discovery strategy. Find more information at http://docs.php-http.org/en/latest/discovery.html#common-errors
>  - Puli Factory is not available
>  - No valid candidate found using strategy "Http\Discovery\Strategy\CommonClassesStrategy". We tested the following candidates: .
>  - No valid candidate found using strategy "Http\Discovery\Strategy\CommonPsr17ClassesStrategy". We tested the following candidates: Nyholm\Psr7\Factory\Psr17Factory, Zend\Diactoros\ResponseFactory, Http\Factory\Diactoros\ResponseFa
  ctory, Http\Factory\Guzzle\ResponseFactory, Http\Factory\Slim\ResponseFactory.

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | Fix #156
| Documentation   |
| License         | MIT

This is a small script to make sure we test different sets of configuration. Related to #156 

Test should fail because #156 describes a problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,20 +23,6 @@ branches:
 matrix:
     fast_finish: true
     include:
-        - name: Test Install
-          php: 7.3
-          install:
-          script:
-              # Test that we find Guzzle
-              - ./tests/install.sh "will-find" "Http\Discovery\HttpClientDiscovery::find();" "php-http/guzzle6-adapter"
-              # Test that we find a client with Symfony and Guzzle PSR-7
-              - ./tests/install.sh will-find "Http\Discovery\HttpClientDiscovery::find();" "symfony/http-client:5.* php-http/httplug php-http/message-factory guzzlehttp/psr7:1.* http-interop/http-factory-guzzle"
-              # We should fail if we dont have php-http/message-factory or PSR-17
-              - ./tests/install.sh cant-find "Http\Discovery\HttpClientDiscovery::find();" "symfony/http-client:5.* php-http/httplug php-http/message-factory guzzlehttp/psr7:1.*"
-              - ./tests/install.sh cant-find "Http\Discovery\HttpClientDiscovery::find();" "symfony/http-client:5.* php-http/httplug guzzlehttp/psr7:1.* http-interop/http-factory-guzzle"
-              # We should be able to find a client when Symfony is only partly installed and we have guzzle adapter installed
-              - ./tests/install.sh will-find "Http\Discovery\HttpClientDiscovery::find();" "symfony/http-client:5.* php-http/guzzle6-adapter php-http/httplug php-http/message-factory guzzlehttp/psr7:1.*"
-
         - name: PHPSpec code coverage
           php: 7.1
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -35,6 +35,38 @@ matrix:
...
+                    (exit $ok) &&
+                        echo -e "\\e[32mOK\\e[0m $title\\n\\ntravis_fold:end:$fold" ||
+                        echo -e "\\e[41mKO\\e[0m $title\\n"
+                    (exit $ok)
+                }
+                export -f install_test
+
+          script:
+              # Test that we find Guzzle
+              - install_test "will-find" "Http\Discovery\HttpClientDiscovery::find();" "php-http/guzzle6-adapter"
```
dbu: quotes are inconsistent with lines below

.travis.yml
```diff
@@ -35,6 +35,38 @@ matrix:
...
+                    echo -e "\\e[1;34m$title\\e[0m"
+                    echo "./tests/install.sh \"$1\" \"$2\" \"$3\""
+                    ./tests/install.sh "$1" "$2" "$3" 2>&1
+                    local ok=$?
+                    (exit $ok) &&
+                        echo -e "\\e[32mOK\\e[0m $title\\n\\ntravis_fold:end:$fold" ||
+                        echo -e "\\e[41mKO\\e[0m $title\\n"
+                    (exit $ok)
+                }
+                export -f install_test
```
dbu: could we add this as a bash file to tests folder, or `.travis` folder? no big deal, but it seems a bit much for the travis configuration file

.travis.yml
Nyholm: Thank you

.travis.yml
Nyholm: I dont think so. Im not sure how you can "export" a function in bash/travis.

Nyholm:
Sure, I added a commit with a fix for #156

Nyholm:
> can we improve output of the build?

What do you suggest?

- Each test is folded so it get easier to see what is what. 
- Each tests ends with a green "OK" or a red "KO" to signal status. 
- The second last line of the tests prints some output that is helpful.

.travis.yml
dbu: ah i see, bash function, not bash script. i guess thats fine then. maybe bash `source`-ing the file would work, but i will merge as is and tag another patch release

dbu:
i somehow expected something *before* the error line, but the last line is enough. as it should normally not fail, it will at least indicate when something is amiss.

Nyholm:
Thank you for merging

Issue #158, opened 2019-12-27
ClassDiscovery::safeClassExists broken


XWB:
Version 1.7.2

It seems https://github.com/php-http/discovery/pull/154 broke something, because I now got the following error:

> No PSR-17 response factory found. Install a package from this list: https://packagist.org/providers/psr/http-factory-implementation

> Could not find resource using any discovery strategy. Find more information at http://docs.php-http.org/en/latest/discovery.html#common-errors - Puli Factory is not available - No valid candidate found using strategy "Http\Discovery\Strategy\CommonClassesStrategy". We tested the following candidates: . - No valid candidate found using strategy "Http\Discovery\Strategy\CommonPsr17ClassesStrategy". We tested the following candidates: Nyholm\Psr7\Factory\Psr17Factory, Zend\Diactoros\ResponseFactory, Http\Factory\Diactoros\ResponseFactory, Http\Factory\Guzzle\ResponseFactory, Http\Factory\Slim\ResponseFactory.

So when I revert this:

```php
    public static function safeClassExists($class)
    {
        try {
            return class_exists($class) || interface_exists($class);
        } catch (\Exception $e) {
            return false;
        }
```

Back to the old behavior:

```php
    public static function safeClassExists($class)
    {
        try {
            return class_exists($class);
        } catch (\Exception $e) {
            return false;
        }
```

Now the error is gone.

The debug trace points to https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/ProxyClient/HttpDispatcher.php#L99

Any idea what this could be?

onEXHovia:
I Confirm the problem on the Version 1.7.3. After install `symfony/http-client` i get the same exception as @XWB.

<img width="1056" alt="Screenshot at Jan 02 21-24-12" src="https://user-images.githubusercontent.com/3332033/71678139-41347780-2da6-11ea-8c0e-c189111393ef.png">

XWB:
Fixed in https://github.com/php-http/discovery/releases/tag/1.7.4

PR #159, opened 2020-01-02
Adding tests for Symfony Httplug Async

XWB:
Fixes https://github.com/php-http/discovery/issues/158

The change introduced in https://github.com/php-http/discovery/commit/366a549f1a9d8f0ac6fc91535a42d7080b560e57 breaks our application.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,6 +65,8 @@ matrix:
               - install_test cant-find "Http\Discovery\HttpClientDiscovery::find();" "symfony/http-client:5.* php-http/httplug guzzlehttp/psr7:1.* http-interop/http-factory-guzzle"
               # We should be able to find a client when Symfony is only partly installed and we have guzzle adapter installed
               - install_test will-find "Http\Discovery\HttpClientDiscovery::find();" "symfony/http-client:5.* php-http/guzzle6-adapter php-http/httplug php-http/message-factory guzzlehttp/psr7:1.*"
+              # Test that we find a client with Symfony and Guzzle
+              - install_test will-find "Http\Discovery\HttpClientDiscovery::find();" "symfony/http-client:5.* php-http/guzzle6-adapter
 
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,9 +66,9 @@ matrix:
               # We should be able to find a client when Symfony is only partly installed and we have guzzle adapter installed
               - install_test will-find "Http\Discovery\HttpClientDiscovery::find();" "symfony/http-client:5.* php-http/guzzle6-adapter php-http/httplug php-http/message-factory guzzlehttp/psr7:1.*"
               # Test that we find a client with Symfony and Guzzle
-              - install_test will-find "Http\Discovery\HttpClientDiscovery::find();" "symfony/http-client:5.* php-http/guzzle6-adapter
+              - install_test will-find "Http\Discovery\HttpClientDiscovery::find();" "symfony/http-client:4.* php-http/guzzle6-adapter
               # Test that we find an async client with Symfony and Guzzle
-              - install_test will-find "Http\Discovery\HttpAsyncClientDiscovery::find();" "symfony/http-client:5.* php-http/guzzle6-adapter
+              - install_test will-find "Http\Discovery\HttpAsyncClientDiscovery::find();" "symfony/http-client:4.* php-http/guzzle6-adapter
 
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,9 +66,9 @@ matrix:
               # We should be able to find a client when Symfony is only partly installed and we have guzzle adapter installed
               - install_test will-find "Http\Discovery\HttpClientDiscovery::find();" "symfony/http-client:5.* php-http/guzzle6-adapter php-http/httplug php-http/message-factory guzzlehttp/psr7:1.*"
               # Test that we find a client with Symfony and Guzzle
-              - install_test will-find "Http\Discovery\HttpClientDiscovery::find();" "symfony/http-client:4.* php-http/guzzle6-adapter
+              - install_test will-find "Http\Discovery\HttpClientDiscovery::find();" "symfony/http-client:4.* php-http/guzzle6-adapter"
               # Test that we find an async client with Symfony and Guzzle
-              - install_test will-find "Http\Discovery\HttpAsyncClientDiscovery::find();" "symfony/http-client:4.* php-http/guzzle6-adapter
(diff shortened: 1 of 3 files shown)


dbu:
i don't think this is the correct fix. @Nyholm made a precondition based on an interface. if the scenario does have that interface available but is still not able to instantiate, we need to add additional criteria for instantiation. 

i guess the problem is that checking for the interface can not tell if any implementation is available. but if there is a whole bunch of possible implementations, i dunno how we can determine if there is one of them present... @Nyholm do you have an idea how to handle it? could we tentatively call some factory to see if it finds a solution?

Nyholm:
I love this library but the past few months it feels like we've done too many patch fixes.. 
A few weeks ago I introduced some additional testing to avoid this kind of issue. 

@XWB Can you have a look here: https://github.com/php-http/discovery/blob/1.7.3/.travis.yml#L59-L67 and provide your own test that represent your application to show what is broken?

XWB:
I''l try to make a test. In the meantime, here's the stack trace:

```php
 at Http\Discovery\Exception\DiscoveryFailedException::create(array(object(PuliUnavailableException), object(NoCandidateFoundException), object(NoCandidateFoundException)))
     (vendor/php-http/discovery/src/ClassDiscovery.php:79)
  at Http\Discovery\ClassDiscovery::findOneByType('Psr\\Http\\Message\\ResponseFactoryInterface')
     (vendor/php-http/discovery/src/Psr17FactoryDiscovery.php:53)
  at Http\Discovery\Psr17FactoryDiscovery::findResponseFactory()
     (vendor/symfony/http-client/HttplugClient.php:79)
  at Symfony\Component\HttpClient\HttplugClient->__construct()
     (vendor/php-http/discovery/src/ClassDiscovery.php:203)
  at Http\Discovery\ClassDiscovery::instantiateClass('Symfony\\Component\\HttpClient\\HttplugClient')
     (vendor/php-http/discovery/src/HttpAsyncClientDiscovery.php:40)
  at Http\Discovery\HttpAsyncClientDiscovery::find()
     (vendor/friendsofsymfony/http-cache/src/ProxyClient/HttpDispatcher.php:99)
```

As you can see, it's trying to initialize `Psr\\Http\\Message\\ResponseFactoryInterface` which doesn't work.

Nyholm:
That is not what I see. =)

I see that you are finding the Symfony HTTPlug client. When instantiating `HttplugClient` we need a `Psr\\Http\\Message\\ResponseFactoryInterface` so we are trying to find one, but fails. 

I thought this test would cover your case: 
https://github.com/php-http/discovery/blob/1.7.3/.travis.yml#L67

Are you using version 1.7.3?

XWB:
I do have 1.7.3 installed:

```
$ composer show | grep discovery

php-http/discovery                  1.7.3      Finds installed HTTPlug implementations and PSR-7 message factories
```

And everything works fine when I revert https://github.com/php-http/discovery/commit/366a549f1a9d8f0ac6fc91535a42d7080b560e57 so the `interface_exists` must break something?

Nyholm:
Please add a test similar to the one I linked. 
It is only one row where you specify your dependencies. It will be interesting to see.

XWB:
@Nyholm Done. I added two tests with `friendsofsymfony/http-cache-bundle` and you can see the error in Travis:

```
PHP Fatal error:  Uncaught Http\Discovery\Exception\DiscoveryFailedException: Could not find resource using any discovery strategy. Find more information at http://docs.php-http.org/en/latest/discovery.html#common-errors
 - Puli Factory is not available
 - No valid candidate found using strategy "Http\Discovery\Strategy\CommonClassesStrategy". We tested the following candidates: .
 - No valid candidate found using strategy "Http\Discovery\Strategy\CommonPsr17ClassesStrategy". We tested the following candidates: Nyholm\Psr7\Factory\Psr17Factory, Zend\Diactoros\ResponseFactory, Http\Factory\Diactoros\ResponseFactory, Http\Factory\Guzzle\ResponseFactory, Http\Factory\Slim\ResponseFactory.
 in /home/travis/build/php-http/discovery/build/504b61b/vendor/php-http/discovery/src/Exception/DiscoveryFailedException.php:41
Stack trace:
#0 /home/travis/build/php-http/discovery/build/504b61b/vendor/php-http/discovery/src/ClassDiscovery.php(79): Http\Discovery\Exception\DiscoveryFailedException::create(Array)
#1 /home/travis/build/php-http/ in /home/travis/build/php-http/discovery/build/504b61b/vendor/php-http/discovery/src/ClassDiscovery.php on line 210
```

onEXHovia:
✋ I don’t know if it will be useful. In project we use packages with versions:
```
"friendsofsymfony/http-cache-bundle": "^2.8"
"php-http/guzzle6-adapter": "^1.1"
"php-http/httplug-bundle": "^1.17"
```

`php-http/discovery ` have version 1.7.3, current set packages work correct, but after install `symfony/http-client@v4.4.2` i get the same exception which mentioned in the [comment](https://github.com/php-http/discovery/issues/158#issuecomment-570259279).

.travis.yml
```diff
@@ -65,6 +65,10 @@ matrix:
               - install_test cant-find "Http\Discovery\HttpClientDiscovery::find();" "symfony/http-client:5.* php-http/httplug guzzlehttp/psr7:1.* http-interop/http-factory-guzzle"
               # We should be able to find a client when Symfony is only partly installed and we have guzzle adapter installed
               - install_test will-find "Http\Discovery\HttpClientDiscovery::find();" "symfony/http-client:5.* php-http/guzzle6-adapter php-http/httplug php-http/message-factory guzzlehttp/psr7:1.*"
+              # Test that we find a client with Symfony and Guzzle
+              - install_test will-find "Http\Discovery\HttpClientDiscovery::find();" "friendsofsymfony/http-cache-bundle:2.4.* symfony/http-client:4.* php-http/guzzle6-adapter"
+              # Test that we find an async client with Symfony and Guzzle
+              - install_test will-find "Http\Discovery\HttpAsyncClientDiscovery::find();" "friendsofsymfony/http-cache-bundle:2.4.* symfony/http-client:4.* php-http/guzzle6-adapter"
```
Nyholm: You can replace `friendsofsymfony/http-cache-bundle:2.4.*` with `php-http/client-common:2.* php-http/message:1.8*`

It is the same effect.

Nyholm:
I also added #161. Both will fix this issue. Which one do we prefer?

XWB:
@Nyholm I manually applied your fix in our application and it still fails :(

image

image

XWB:
But https://github.com/php-http/discovery/pull/161 fixes the issue! :)

Nyholm:
Fixed and released https://github.com/php-http/discovery/releases/tag/1.7.4

PR #160, opened 2020-01-02
Add PHP 7.4

XWB:
Replaces https://github.com/php-http/discovery/pull/153


XWB:
Looks like this won't work without https://github.com/akeneo/PhpSpecSkipExampleExtension/pull/22

dbu:
argh. can we allow phpspec 5 and 6 in parallel, or is what we do with phpspec affected by bc breaks?

PR #161, opened 2020-01-02
Alternative to Symfony Httplug Async

Nyholm:
This is a alternative to #159 

Im not sure which one I like better. 

This will check if a PSR17Factory can be found before accepting Symfony's HttplugClient as a candidate.

### DIFF
--- a/src/Strategy/CommonClassesStrategy.php
+++ b/src/Strategy/CommonClassesStrategy.php
@@ -69,13 +69,13 @@ final class CommonClassesStrategy implements DiscoveryStrategy
             ['class' => SlimUriFactory::class, 'condition' => [SlimRequest::class, SlimUriFactory::class]],
         ],
         HttpAsyncClient::class => [
-            ['class' => SymfonyHttplug::class, 'condition' => [SymfonyHttplug::class, Promise::class, RequestFactory::class, [CommonClassesStrategy::class, 'isPsr17FactoryInstalled']]],
+            ['class' => SymfonyHttplug::class, 'condition' => [SymfonyHttplug::class, Promise::class, RequestFactory::class, [self::class, 'isPsr17FactoryInstalled']]],
             ['class' => Guzzle6::class, 'condition' => Guzzle6::class],
             ['class' => Curl::class, 'condition' => Curl::class],
             ['class' => React::class, 'condition' => React::class],
         ],
         HttpClient::class => [
-            ['class' => SymfonyHttplug::class, 'condition' => [SymfonyHttplug::class, RequestFactory::class, [CommonClassesStrategy::class, 'isPsr17FactoryInstalled']]],
+            ['class' => SymfonyHttplug::class, 'condition' => [SymfonyHttplug::class, RequestFactory::class, [self::class, 'isPsr17FactoryInstalled']]],
             ['class' => Guzzle6::class, 'condition' => Guzzle6::class],
             ['class' => Guzzle5::class, 'condition' => Guzzle5::class],
             ['class' => Curl::class, 'condition' => Curl::class],
@@ -139,6 +139,7 @@ final class CommonClassesStrategy implements DiscoveryStrategy
 
     /**
      * Can be used as a condition.
+     *
      * @return bool
      */
     public static function isPsr17FactoryInstalled()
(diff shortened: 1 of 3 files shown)


src/Strategy/CommonClassesStrategy.php
```diff
@@ -68,13 +69,13 @@ final class CommonClassesStrategy implements DiscoveryStrategy
             ['class' => SlimUriFactory::class, 'condition' => [SlimRequest::class, SlimUriFactory::class]],
         ],
         HttpAsyncClient::class => [
-            ['class' => SymfonyHttplug::class, 'condition' => [SymfonyHttplug::class, Promise::class, RequestFactory::class]],
+            ['class' => SymfonyHttplug::class, 'condition' => [SymfonyHttplug::class, Promise::class, RequestFactory::class, [CommonClassesStrategy::class, 'isPsr17FactoryInstalled']]],
```
XWB: How about `self::class` instead of `CommonClassesStrategy::class`?

onEXHovia:
@Nyholm thanks, it's works for me

XWB:
This change works for me as well.

But the tests still fail.

src/Strategy/CommonClassesStrategy.php
```diff
@@ -135,4 +136,19 @@ public static function symfonyPsr18Instantiate()
     {
         return new SymfonyPsr18(null, Psr17FactoryDiscovery::findResponseFactory(), Psr17FactoryDiscovery::findStreamFactory());
     }
+
+    /**
+     * Can be use as a condition.
```
xabbuh: Can be used [...]

src/Strategy/CommonClassesStrategy.php
```diff
@@ -135,4 +136,19 @@ public static function symfonyPsr18Instantiate()
...
+     * @return bool
+     */
+    public static function isPsr17FactoryInstalled()
+    {
+        try {
+            Psr17FactoryDiscovery::findResponseFactory();
+        } catch (NotFoundException $e) {
```
dbu: could there be other exceptions, e.g. when a class can't be loaded because missing interface or whatnot?

dbu: maybe we should catch any throwable

Nyholm: It can be other exceptions.. but are we interested in those? 
Maybe we should do a warning then

src/Strategy/CommonClassesStrategy.php
Nyholm: What is the difference?

src/Strategy/CommonClassesStrategy.php
```diff
@@ -135,4 +136,23 @@ public static function symfonyPsr18Instantiate()
...
+            return false;
+        } catch (\Throwable $e) {
+            trigger_error('Trying to find a PSR-17 ResponseFactory when an exception got thrown: '.$e->getMessage(), E_USER_ERROR);
```
Nyholm: Please help me here, should it be `E_USER_ERROR`? Should it be silenced with a `@`?

src/Strategy/CommonClassesStrategy.php
Nyholm: It looks like the style guide agrees with you =)

src/Strategy/CommonClassesStrategy.php
Nyholm: Thank you

src/Strategy/CommonClassesStrategy.php
dbu: i would make it a E_USER_WARNING `sprintf('Got exception "%s (%s)" while checking if a PSR-17 ResponseFactory is available', get_class($e), $e->getMessage())`

dbu: if we assume that discovery should try more things in that case. if the psr-17 factory was expected to be viable, then "hiding" the problem in a E_USER_WARNING is not helpful, but if something else is viable and psr-17 is only sort of detected but not viable, the warning makes it so we can continue.

Nyholm: > if we assume that discovery should try more things in that case. 

We do.

Nyholm:
Wohoo, the "install tests" are useful. I found a bug =)

We still check for interfaces. Ie, Symfony client does not work without the `Http\Message\RequestFactory` (interface)

Nyholm:
Thank you

PR #167, opened 2020-03-06
Support discovering PSR-17 factories from guzzlehttp/psr7 package

### LINKED ISSUE #164: "No PSR-17 request factory found" incorrectly suggests guzzlehttp/psr7
**PHP version**: 7.3

**Description**
Hi!

This seems related to #156 and #149.

I received the error: 

> No PSR-17 request factory found. Install a package from this list: https://packagist.org/providers/psr/http-factory-implementation

But when I go to https://packagist.org/providers/psr/http-factory-implementation, the first item on the list is `guzzlehttp/psr7`. When I installed it, I got the same error. After some deep digging, If found out that I actually needed `http-interop/http-factory-guzzle`. Installing that package *does* fix things - it's just the recommendation that was wrong (and tough to debug!).

Am I correct that `guzzlehttp/psr7` should not be shown on the list of "http factory implementations"? If so, why is it there? Is it a bug in *that* package's metadata? Something on Packagist? I don't know much about this - I'm just trying to get a bundle that is a few layers above this to work :).

Thanks!

ste93cry:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | it depends on the outcome of the discussion
| Deprecations?   | no
| Related tickets | fixes #164
| License         | MIT

#### What's in this PR?

In the next version of [`guzzlehttp/psr7`](https://github.com/guzzle/psr7) there will be support for PSR-17 (actually the feature has not been released yet), so I simply added support for discovering the factories.

#### Checklist

#### To Do

For the moment I added the new factories in the middle of the list instead of at the end. What this means is that suddenly the priorities of discovery will change and the `guzzle/psr7` package will be preferred over the `php-http/guzzle6-adapter`, which may be considered a BC. Just let me know if I have to move the classes to the end of the list instead

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,16 @@
 # Change Log
 
+## Unreleased
+
+### Added
+
+- Support discovering PSR-17 factories of `guzzlehttp/psr7` package
+
 ## 1.7.4 - 2020-01-03
 
 ### Fixed
 
-- Improve conditions on Symfony's async HTTPlug client. 
+- Improve conditions on Symfony's async HTTPlug client.
 
 ## 1.7.3 - 2019-12-27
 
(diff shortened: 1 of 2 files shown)


ste93cry:
@Nyholm @dbu is there anything left to do or can this be merged?

PR #175, opened 2020-06-28
Support Guzzle 7+

### LINKED ISSUE #174: Guzzle 7 Not Found
**PHP version**: 7.4.7

**Description**

Cannot find Guzzle 7.0.1, even though the package provides a psr18 implementation (https://github.com/guzzle/guzzle/blob/7.0.1/composer.json#L38).

**How to reproduce**

`composer show`:

```
guzzlehttp/guzzle                  7.0.1   Guzzle is a PHP HTTP client library
guzzlehttp/promises                v1.3.1  Guzzle promises library
guzzlehttp/psr7                    1.6.1   PSR-7 message implementation tha...
php-http/cache-plugin              1.7.0   PSR-6 Cache plugin for HTTPlug
php-http/client-common             2.1.0   Common HTTP Client implementatio...
php-http/discovery                 1.8.0   Finds installed HTTPlug implemen...
php-http/httplug                   2.1.0   HTTPlug, the HTTP client abstrac...
php-http/message                   1.8.0   HTTP Message related tools
php-http/message-factory           v1.0.2  Factory interfaces for PSR-7 HTT...
php-http/multipart-stream-builder  1.1.0   A builder class that help you cr...
php-http/promise                   v1.0.0  Promise used for asynchronous HT...
```

Example code:

```php
Psr18ClientDiscovery::find();
```

Error:

```
Http\Discovery\Exception\NotFoundException: No PSR-18 clients found. Make sure to install a package providing "psr/http-client-implementation". Example: "php-http/guzzle6-adapter".

/data/vendor/php-http/discovery/src/Psr18ClientDiscovery.php:27
/data/src/HttpClient/Builder.php:108
/data/src/Client.php:79
/data/tests/ClientTest.php:29

Caused by
Http\Discovery\Exception\DiscoveryFailedException: Could not find resource using any discovery strategy. Find more information at http://docs.php-http.org/en/latest/discovery.html#common-errors
 - Puli Factory is not available
 - No valid candidate found using strategy "Http\Discovery\Strategy\CommonClassesStrategy". We tested the following candidates: Http\Discovery\Strategy\CommonClassesStrategy::symfonyPsr18Instantiate, Http\Discovery\Strategy\CommonClassesStrategy::buzzInstantiate.
 - No valid candidate found using strategy "Http\Discovery\Strategy\CommonPsr17ClassesStrategy". We tested the following candidates: .



/data/vendor/php-http/discovery/src/Exception/DiscoveryFailedException.php:41
/data/vendor/php-http/discovery/src/ClassDiscovery.php:79
/data/vendor/php-http/discovery/src/Psr18ClientDiscovery.php:25
```

GrahamCampbell:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #174
| License         | MIT

#### What's in this PR?

Adds support for Guzzle 7 and higher.

#### Checklist

#### To Do

### DIFF
--- a/src/Strategy/CommonClassesStrategy.php
+++ b/src/Strategy/CommonClassesStrategy.php
@@ -114,7 +114,7 @@ final class CommonClassesStrategy implements DiscoveryStrategy
     {
         $candidates = [];
 
-        // Not all versions of all clients implement the same interfaces. For example, 
+        // Not all versions of all clients implement the same interfaces. For example,
         // Guzzle 6 does not implement the PSR-18 client interface, but Guzzle 7 does.
         foreach (self::$classes[$type] ?? [] as $c) {
             try {
--- a/src/Strategy/CommonClassesStrategy.php
+++ b/src/Strategy/CommonClassesStrategy.php
@@ -116,7 +116,7 @@ final class CommonClassesStrategy implements DiscoveryStrategy
 
         foreach (self::$classes[$type] ?? [] as $c) {
             // Guzzle 6 does not implement the PSR-18 client interface, but Guzzle 7 does.
-            if (Psr18Client::class === $type && $c['class'] === GuzzleHttp::class) {
+            if (Psr18Client::class === $type && GuzzleHttp::class === $c['class']) {
                 if (defined('GuzzleHttp\ClientInterface::MAJOR_VERSION')) {
                     $candidates[] = $c;
                 }
--- a/src/Strategy/CommonClassesStrategy.php
+++ b/src/Strategy/CommonClassesStrategy.php
@@ -114,15 +114,14 @@ final class CommonClassesStrategy implements DiscoveryStrategy
     {
         $candidates = [];
 
-        // Not all versions of all clients implement the same interfaces. For example,
-        // Guzzle 6 does not implement the PSR-18 client interface, but Guzzle 7 does.
         foreach (self::$classes[$type] ?? [] as $c) {
-            try {
-                if (is_subclass_of($c['class'], $type)) {
+            // Guzzle 6 does not implement the PSR-18 client interface, but Guzzle 7 does.
+            if (Psr18Client::class === $type && $c['class'] === GuzzleHttp::class) {
+                if (defined('GuzzleHttp\ClientInterface::MAJOR_VERSION')) {
                     $candidates[] = $c;
                 }
-            } catch (\Throwable $e) {
-                trigger_error(sprintf('Got exception "%s (%s)" while checking if a PSR-18 Client is available', get_class($e), $e->getMessage()), E_USER_WARNING);
+            } else {
+                $candidates[] = $c;
             }
         }
 
--- a/src/Strategy/CommonClassesStrategy.php
+++ b/src/Strategy/CommonClassesStrategy.php
@@ -26,6 +26,7 @@ use Http\Message\MessageFactory\SlimMessageFactory;
 use Http\Message\StreamFactory\SlimStreamFactory;
 use Http\Message\UriFactory\SlimUriFactory;
 use Slim\Http\Request as SlimRequest;
+use GuzzleHttp\Client as GuzzleHttp;
 use Http\Adapter\Guzzle6\Client as Guzzle6;
 use Http\Adapter\Guzzle5\Client as Guzzle5;
 use Http\Client\Curl\Client as Curl;
@@ -76,6 +77,7 @@ final class CommonClassesStrategy implements DiscoveryStrategy
         ],
         HttpClient::class => [
             ['class' => SymfonyHttplug::class, 'condition' => [SymfonyHttplug::class, RequestFactory::class, [self::class, 'isPsr17FactoryInstalled']]],
+            ['class' => GuzzleHttp::class, 'condition' => GuzzleHttp::class],
             ['class' => Guzzle6::class, 'condition' => Guzzle6::class],
             ['class' => Guzzle5::class, 'condition' => Guzzle5::class],
             ['class' => Curl::class, 'condition' => Curl::class],
--- a/src/Strategy/CommonClassesStrategy.php
+++ b/src/Strategy/CommonClassesStrategy.php
@@ -77,7 +77,6 @@ final class CommonClassesStrategy implements DiscoveryStrategy
         ],
         HttpClient::class => [
             ['class' => SymfonyHttplug::class, 'condition' => [SymfonyHttplug::class, RequestFactory::class, [self::class, 'isPsr17FactoryInstalled']]],
-            ['class' => GuzzleHttp::class, 'condition' => GuzzleHttp::class],
             ['class' => Guzzle6::class, 'condition' => Guzzle6::class],
             ['class' => Guzzle5::class, 'condition' => Guzzle5::class],
             ['class' => Curl::class, 'condition' => Curl::class],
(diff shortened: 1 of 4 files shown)


src/Strategy/CommonClassesStrategy.php
```diff
@@ -76,6 +77,7 @@ final class CommonClassesStrategy implements DiscoveryStrategy
         ],
         HttpClient::class => [
             ['class' => SymfonyHttplug::class, 'condition' => [SymfonyHttplug::class, RequestFactory::class, [self::class, 'isPsr17FactoryInstalled']]],
+            ['class' => GuzzleHttp::class, 'condition' => GuzzleHttp::class],
```
gmponos, line 80: I think this is not the correct place.. this is a list with clients that implement the httplug interface..

gmponos: I think this is the place where we want guzzle: https://github.com/php-http/discovery/pull/175/files#diff-f6115264dc0232837d2b0b916ccb41b8R95

GrahamCampbell:
PHPUnit tests seem to be failing.

Nyholm:
Why not improve the condition to make sure the correct version of Guzzle is used?

GrahamCampbell:
I think my changes have actually exposed another bug, have they not? Those failing tests should never have passed?

GrahamCampbell:
> Why not improve the condition to make sure the correct version of Guzzle is used?

Changing conditions to be able to do more than check class existance would be a breaking change?

src/Strategy/CommonClassesStrategy.php
```diff
@@ -120,15 +136,9 @@ public static function getCandidates($type)
                     trigger_error(sprintf('Got exception "%s (%s)" while checking if a PSR-18 Client is available', get_class($e), $e->getMessage()), E_USER_WARNING);
                 }
             }
-
-            return $candidates;
-        }
-
-        if (isset(self::$classes[$type])) {
-            return self::$classes[$type];
```
dbu, line 128: don't we still need this bit?

GrahamCampbell: No, it was replaced by the the foreach at the top of the method.

src/Strategy/CommonClassesStrategy.php
```diff
@@ -107,9 +112,20 @@ final class CommonClassesStrategy implements DiscoveryStrategy
      */
     public static function getCandidates($type)
     {
-        if (Psr18Client::class === $type) {
-            $candidates = self::$classes[PSR18Client::class];
+        $candidates = [];
 
+        foreach (self::$classes[$type] ?? [] as $c) {
+            // Guzzle 6 does not implement the PSR-18 client interface, but Guzzle 7 does.
+            if (Psr18Client::class === $type && GuzzleHttp::class === $c['class']) {
```
dbu, line 119: i find this hard to read.

what if we would do

```
$candidates = self::$classes[$type] ?? [];
if (Psr18Client::class === $type) {
    if ( !defined(GuzzleHttp\ClientInterface::class.'::MAJOR_VERSION') && $pos = array_search(GuzzleHttp::class, $candidates)) {
        unset($candidates[$pos]);
    }
    foreach (self::$classes[HttpClient::class] as $c) {
        ...
    }
}
```

that way, the special case of psr18 with guzzle is more clear. we don't do the if in each iteration, and don't loop when we are not about psr18 client.

dbu: or even `if (Psr18Client::class !== $type) return $c;` for an early return.

GrahamCampbell: We don't want an early return. I think what I wrote is the most readable out of everything there. I think it's pretty clear. We start with an empty array of candidates, and then we add in the ones we want, then return at the bottom of the function, in all cases.

dbu: sorry, i want to continue challenging this bit of code a bit.

i find it difficult to see what exactly we do, and part of the `if` condition inside the loop is not depending on anything in the loop. for non psr18 client we do not need to loop at all.

how about this:

```
if (Psr18Client::class !== $type) {
    return self::$classes[$type] ?? [];;
}

$candidates = [];
foreach (self::$classes[Psr18Client::class] as $c) {
    if (GuzzleHttp::class !== $c || defined(GuzzleHttp\ClientInterface::class.'::MAJOR_VERSION')) {
        $candidates[] = $c;
    }
    // HTTPlug 2.0 clients implements PSR18Client too.
    foreach (self::$classes[HttpClient::class] as $c) {
        ...
    }
}
```

GrahamCampbell: I've updated it again, to handle PSR-18 stuff totally separately, to avoid confusion. Is that more to your taste now?

```php

    /**
     * {@inheritdoc}
     */
    public static function getCandidates($type)
    {
        if (Psr18Client::class === $type) {
            return self::getPsr18Candidates();
        }

        return self::$classes[$type] ?? [];
    }

    /**
     * @return array The return value is always an array with zero or more elements. Each
     *               element is an array with two keys ['class' => string, 'condition' => mixed].
     */
    private static function getPsr18Candidates()
    {
        $candidates = [];

        // Guzzle 6 does not implement the PSR-18 client interface, but Guzzle 7 does.
        foreach (self::$classes[Psr18Client::class] ?? [] as $c) {
            if (GuzzleHttp::class === $c['class']) {
                if (defined('GuzzleHttp\ClientInterface::MAJOR_VERSION')) {
                    $candidates[] = $c;
                }
            } else {
                $candidates[] = $c;
            }
        }

        // HTTPlug 2.0 clients implements PSR18Client too.
        foreach (self::$classes[HttpClient::class] as $c) {
            try {
                if (is_subclass_of($c['class'], Psr18Client::class)) {
                    $candidates[] = $c;
                }
            } catch (\Throwable $e) {
                trigger_error(sprintf('Got exception "%s (%s)" while checking if a PSR-18 Client is available', get_class($e), $e->getMessage()), E_USER_WARNING);
            }
        }

        return $candidates;
    }
```

GrahamCampbell:
Ready for re-review now. I've also expanded the "install" tests, so we have some proof that we can detect Guzzle 6 and 7 as PSR-18 clients, and that this won't break in the future.

src/Strategy/CommonClassesStrategy.php
```diff
@@ -108,27 +113,43 @@ final class CommonClassesStrategy implements DiscoveryStrategy
...
-                try {
-                    if (is_subclass_of($c['class'], Psr18Client::class)) {
-                        $candidates[] = $c;
-                    }
-                } catch (\Throwable $e) {
-                    trigger_error(sprintf('Got exception "%s (%s)" while checking if a PSR-18 Client is available', get_class($e), $e->getMessage()), E_USER_WARNING);
+        // Guzzle 6 does not implement the PSR-18 client interface, but Guzzle 7 does.
+        foreach (self::$classes[Psr18Client::class] ?? [] as $c) {
+            if (GuzzleHttp::class === $c['class']) {
+                if (defined('GuzzleHttp\ClientInterface::MAJOR_VERSION')) {
```
Nyholm, line 133: I dont think we need this. 

This logic should be on the `condition`

Nyholm:
Fixed in https://github.com/php-http/discovery/pull/177

PR #177, opened 2020-07-02
Refactor to remove "guzzle specific" logic

Nyholm:
Follow up from #175

FYI @GrahamCampbell and @dbu


dbu:
indeed, this is more compact 

i don't think we need a new release for this, as its functionally doing the same. agreed?

PR #180, opened 2020-07-02
PHP 7.4 and 8.0 support

GrahamCampbell:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #160
| License         | MIT

### DIFF
--- a/spec/autoload.php
+++ b/spec/autoload.php
@@ -8,8 +8,6 @@ if (PHP_VERSION_ID >= 50600) {
     ]);
 }
 
-if (PHP_VERSION_ID >= 70300 && PHP_VERSION_ID < 70400) {
-    class PHP73 {}
-} else {
-    class NotPHP73 {}
+if (PHP_VERSION_ID < 70300) {
+    class PuliIsSupported {}
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,8 @@
 language: php
 
-sudo: false
-
 cache:
     directories:
         - $HOME/.composer/cache/files
-
-php:
-    - 7.1
-    - 7.2
-    - 7.3
 
 env:
     global:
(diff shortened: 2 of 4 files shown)


GrahamCampbell:
Welp, this technically works :trollface:. Avoids the problem of the skipping package not being maintained.

spec/autoload.php
```diff
@@ -8,7 +8,7 @@
     ]);
 }
 
-if (PHP_VERSION_ID >= 70300 && PHP_VERSION_ID < 70400) {
+if (PHP_VERSION_ID >= 70300) {
```
dbu, line 11: this was a bit odd. @Nyholm would you remember if this was done for a specific reason?

if i understand correctly, this is only for the tests, so we could rename the classes without isuses. if the semantic of the if is correct, AtLeastPHP73 and LessThanPHP73 would be better names.

if we would ever need an upper bound, we can introduce it and rename the classes again.

GrahamCampbell: At the time Puli didn;t work with PHP 7.3 yet, and that was the latest version. 2 years later, we are now in the siutation where 7.4 and 8.0 don't work either. I guess I can rename this class if we want. Let me know what name you prefer. :)

GrahamCampbell: As for what this class does, yes, it exists so we can skip the puli tests.

GrahamCampbell:
@dbu This is ready now. :)

spec/autoload.php
dbu: yep. lets call them AtLeastPHP73 and LessThanPHP73? or possibly: PuliCompatible and PuliIncompatible (if we even need both and not only one)

GrahamCampbell: Already renamed it around an hour ago btw. :)

PR #184, opened 2020-09-03
update styleci configuration

dbu:


dbu:
@sagikazarmark do you have any idea what is wrong with phpspec 8? https://travis-ci.org/github/php-http/discovery/jobs/723682634

from what i can see everything looks to be in the right place for PSR-4. but spec has some PSR-0 thing complaining... do we need to configure something additional for spec version 8 that was not necessary in 7?

PR #187, opened 2020-09-21
1.10.1 release

joelwurtz:
Ok to release ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## 1.10.1 - 2020-09-21
+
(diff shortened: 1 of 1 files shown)


dbu (approved):
i am a bit worried by the phpspec failure but it seems to be about some change with phpspec in version 8 and not about our code.

i am ok with releasing as is. if somebody has time to fix the phpspec 8 build, it would be cool, but can be done after the release too.

GrahamCampbell:
Should be a 1.11.0, since a new deprecation cannot be in a patch release?

joelwurtz:
fixed in #188

PR #195, opened 2021-03-04
Do not suggest installation of puli via composer or error messages

naderman:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | 
| Documentation   | 
| License         | MIT

#### What's in this PR?

Puli has not been modified in 5 years, it's abandoned, the website is down and it's not compatible with Composer v2, it should be removed from this project.

More urgently it is currently the first suggestion for why discovery may not work, leading novice users to install puli to fix this error message, which just makes things worse.

So this PR for now just removes the suggestion from composer.json and the error message.

Next step would be proper full deprecation and eventually removal.

#### Checklist

- [x] Documentation pull request created (if not simply a bugfix)

### DIFF
--- a/src/ClassDiscovery.php
+++ b/src/ClassDiscovery.php
@@ -22,9 +22,9 @@ abstract class ClassDiscovery
      * @var array
      */
     private static $strategies = [
-        Strategy\PuliBetaStrategy::class,
         Strategy\CommonClassesStrategy::class,
         Strategy\CommonPsr17ClassesStrategy::class,
+        Strategy\PuliBetaStrategy::class,
     ];
 
     private static $deprecatedStrategies = [
--- a/src/ClassDiscovery.php
+++ b/src/ClassDiscovery.php
@@ -27,6 +27,10 @@ abstract class ClassDiscovery
         Strategy\CommonPsr17ClassesStrategy::class,
     ];
 
+    private static $deprecatedStrategies = [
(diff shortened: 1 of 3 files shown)


dbu (approved):
seems reasonable to me. @Nyholm any objections?

src/ClassDiscovery.php
```diff
@@ -54,8 +58,10 @@ protected static function findOneByType($type)
         foreach (self::$strategies as $strategy) {
             try {
                 $candidates = call_user_func($strategy.'::getCandidates', $type);
-            } catch (StrategyUnavailableException $e) {
-                $exceptions[] = $e;
+	    } catch (StrategyUnavailableException $e) {
```
Nyholm, line 61: Please increase indentation.

Nyholm (commented):
Yeah: I agree with this. 

You can also change the order or `ClassDiscovery::$strategies`

src/ClassDiscovery.php
naderman: Ugh that's using tabs. New machine, editor not configured properly yet XD

Nyholm:
Thank you

dbu:
i added a note to the changelog for this https://github.com/php-http/discovery/pull/196

GrahamCampbell:
Sorry to see that project die. :(

PR #199, opened 2021-09-16
Add request alias specific to the Laminas namespace

### LINKED ISSUE #198: Diactoros Detection Fails after laminas release 2.6
Because laminas/diactoros now conflicts with zend/diactoros, and the zend framework bridge component is getting deprecated and removed from most Laminas repos, discovery fails if you're using recent laminas versions.

This could be solved either by adding the Laminas equivalents to common classes, or by adding the zend framework bridge component to composer.json.

Assuming the former would be preferable, I'll try and get a pull together shortly.

gsteel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| Related tickets | fixes #198
| License         | MIT

#### What's in this PR?

Since [laminas-diactoros 2.7.0](https://github.com/laminas/laminas-diactoros/releases/tag/2.7.0) - it appears that Message Factory discovery fails _(When the deprecated `MessageFactoryDiscovery` class is used)_

Pretty sure that this is down to the removal of [laminas-zendframework-bridge](https://github.com/laminas/laminas-zendframework-bridge) in that release.

#### To reproduce

```shell
composer require php-http/discovery
composer require laminas/laminas-diactoros ^2.7
composer require php-http/message-factory
```

And then run a php script along the lines of:

```php
<?php
require 'vendor/autoload.php';

use Http\Discovery\MessageFactoryDiscovery;
$factory = MessageFactoryDiscovery::find();

```

This should yield a DiscoveryFailed exception.

Apologies, but I'm not sure how to go about writing tests for this scenario…

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## 1.14.1 - TBC
+
+- [#199](https://github.com/php-http/discovery/pull/199) - Fixes message factory discovery for `laminas-diactoros ^2.7` 
+
 ## 1.14.0 - 2021-06-21
 
 - Deprecate puli as it has been unmaintained for a long time and is not compatible with composer 2 https://github.com/php-http/discovery/pull/195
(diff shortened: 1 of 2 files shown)


dbu (approved):
thanks for this fix!

dbu:
https://github.com/php-http/discovery/releases/tag/1.14.1

Issue #200, opened 2021-09-18
Convert CI build from travis-ci to github actions


dbu:
we currently have a .travis.yml file in this repository, but travis-ci does no longer run it. we should convert to github actions.

PR #202, opened 2021-11-11
symfony httplug adapter needs php-http httplug to be installed

### LINKED ISSUE #201: Undocumented error triggered when using Discovery with Symfony when php-http/httplug not installed
I am using `Psr18ClientDiscovery::find()` in a library I'm writing, and I wrap the `find()` call in a try/catch to capture `\Http\Discovery\Exception\NotFoundException`. I used the library to do a manual test in a Symfony application, but instead of seeing any exception, I see the following `E_USER_WARNING` is raised:

```
User Warning: Got exception "LogicException (You cannot use "Symfony\Component\HttpClient\HttplugClient" as the "php-http/httplug" package is not installed. Try running "composer require php-http/httplug".)" while checking if a PSR-18 Client is available
```

The message itself is fine, but I expect the library to throw a `\Http\Discovery\Exception\NotFoundException`, not trigger an error. I traced the change into the change #170 which catches [the `\LogicException` thrown in Symfony](https://github.com/symfony/http-client/blob/290eb481973b4984eb59377585f6afbc65a0f645/HttplugClient.php#L43-L49), but instead of letting the exception bubble up, or wrapping it in a `\Http\Discovery\Exception\NotFoundException`, it converts it to a runtime error using `trigger_error` which was not expected or documented:

https://github.com/php-http/discovery/blob/de90ab2b41d7d61609f504e031339776bc8c7223/src/Psr18ClientDiscovery.php#L15-L22

As a workaround, one can catch `\ErrorException` as well as `\Http\Discovery\Exception\NotFoundException` (or just catch `\Throwable`), but the fact remains that this is confusing and undocumented behaviour. 

I think rather than converting the `\LogicException` into a `trigger_error`, the exception should be caught and converted to outwardly throw a `NotFoundException`, which is documented behaviour.

dbu:
if we don't check for the httplug client interface, we trigger a warning:
https://github.com/symfony/http-client/blob/290eb481973b4984eb59377585f6afbc65a0f645/HttplugClient.php#L44

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #201
| Documentation   | -
| License         | MIT

#### What's in this PR?

avoid having symfony trigger a warning when symfony httplug client is not usable due to missing dependencies.

### DIFF
--- a/src/Psr18ClientDiscovery.php
+++ b/src/Psr18ClientDiscovery.php
@@ -24,7 +24,7 @@ final class Psr18ClientDiscovery extends ClassDiscovery
         try {
             $client = static::findOneByType(ClientInterface::class);
         } catch (DiscoveryFailedException $e) {
-            throw new \Http\Discovery\Exception\NotFoundException('No PSR-18 clients found. Make sure to install a package providing "psr/http-client-implementation". Example: "php-http/guzzle6-adapter".', 0, $e);
+            throw new \Http\Discovery\Exception\NotFoundException('No PSR-18 clients found. Make sure to install a package providing "psr/http-client-implementation". Example: "php-http/guzzle7-adapter".', 0, $e);
         }
 
         return static::instantiateClass($client);
--- a/src/Strategy/CommonClassesStrategy.php
+++ b/src/Strategy/CommonClassesStrategy.php
@@ -17,6 +17,7 @@ use Http\Client\Curl\Client as Curl;
 use Http\Client\HttpAsyncClient;
 use Http\Client\HttpClient;
 use Http\Client\Socket\Client as Socket;
+use Http\Discovery\ClassDiscovery;
 use Http\Discovery\Exception\NotFoundException;
 use Http\Discovery\MessageFactoryDiscovery;
 use Http\Discovery\Psr17FactoryDiscovery;
@@ -136,8 +137,11 @@ final class CommonClassesStrategy implements DiscoveryStrategy
 
         // HTTPlug 2.0 clients implements PSR18Client too.
         foreach (self::$classes[HttpClient::class] as $c) {
+            if (!is_string($c['class'])) {
+                continue;
+            }
             try {
-                if (is_subclass_of($c['class'], Psr18Client::class)) {
+                if (ClassDiscovery::safeClassExists($c['class']) && is_subclass_of($c['class'], Psr18Client::class)) {
                     $candidates[] = $c;
                 }
             } catch (\Throwable $e) {


dbu:
i don't have time to do #200 - we should convert from travis-ci to github actions to have the tests actually run in this repository.

src/Psr18ClientDiscovery.php
```diff
@@ -24,7 +24,7 @@ public static function find()
         try {
             $client = static::findOneByType(ClientInterface::class);
         } catch (DiscoveryFailedException $e) {
-            throw new \Http\Discovery\Exception\NotFoundException('No PSR-18 clients found. Make sure to install a package providing "psr/http-client-implementation". Example: "php-http/guzzle6-adapter".', 0, $e);
+            throw new \Http\Discovery\Exception\NotFoundException('No PSR-18 clients found. Make sure to install a package providing "psr/http-client-implementation". Example: "php-http/guzzle7-adapter".', 0, $e);
```
asgrim, line 27: Just a query here on deps; `php-http/guzzle6-adapter` is compatible with PHP 7.1 (as `php-http/discovery` is itself), but `php-http/guzzle7-adapter` isn't. This change would now recommend `php-http/guzzle7-adapter` when PHP 7.1 users cannot actually install this?

src/Strategy/CommonClassesStrategy.php
```diff
@@ -75,14 +75,14 @@ final class CommonClassesStrategy implements DiscoveryStrategy
...
-            ['class' => SymfonyHttplug::class, 'condition' => [SymfonyHttplug::class, Promise::class, RequestFactory::class, [self::class, 'isPsr17FactoryInstalled']]],
+            ['class' => SymfonyHttplug::class, 'condition' => [SymfonyHttplug::class, HttpClient::class, Promise::class, RequestFactory::class, [self::class, 'isPsr17FactoryInstalled']]],
             ['class' => Guzzle7::class, 'condition' => Guzzle7::class],
             ['class' => Guzzle6::class, 'condition' => Guzzle6::class],
             ['class' => Curl::class, 'condition' => Curl::class],
             ['class' => React::class, 'condition' => React::class],
         ],
         HttpClient::class => [
-            ['class' => SymfonyHttplug::class, 'condition' => [SymfonyHttplug::class, RequestFactory::class, [self::class, 'isPsr17FactoryInstalled']]],
+            ['class' => SymfonyHttplug::class, 'condition' => [SymfonyHttplug::class, HttpClient::class, RequestFactory::class, [self::class, 'isPsr17FactoryInstalled']]],
```
asgrim, line 85: Unfortunately this does not seem to change anything. [Symfony throws a `\LogicException`](https://github.com/symfony/http-client/blob/290eb481973b4984eb59377585f6afbc65a0f645/HttplugClient.php#L43-L49) outside of the class structure, so as soon as the file is loaded, and `php-http/httplug` is missing, the `\LogicException` is thrown.

Then in `\Http\Discovery\Strategy\CommonClassesStrategy::getPsr18Candidates`, that `\LogicException` is caught, and converted to a `trigger_error`:

https://github.com/php-http/discovery/blob/master/src/Strategy/CommonClassesStrategy.php#L139-L145

In trying to debug this, the only ways I found this was working was to turn that try/catch into silence (not ideal...):

```
            try {
                if (is_subclass_of($c['class'], Psr18Client::class)) {
                    $candidates[] = $c;
                }
            } catch (\Throwable $e) {
            }
```

but ALSO you would need to add `HttpClient::class` as a condition to `symfonyPsr18Instantiate` under the `Psr18Client::class` definitions:

```
        Psr18Client::class => [
            [
                'class' => [self::class, 'symfonyPsr18Instantiate'],
                'condition' => [HttpClient::class, SymfonyPsr18::class, Psr17RequestFactory::class],
            ],
```

Silencing the exception feels very wrong, but with a clear comment explaining why, it might be OK, as it doesn't look like there is a logger around to log a warning to.

src/Psr18ClientDiscovery.php
dbu: it is not related to the fix, just stumbled over it when reading the code. it is an example, not a recommendation.

src/Strategy/CommonClassesStrategy.php
dbu: right the problem occurs before conditions are even checked. i realized that we need to use our safe class exists here as well. changed to use that and in my small reproducer that makes the trigger_error go away.

asgrim:
Thank you @dbu !

dbu:
https://github.com/php-http/discovery/releases/tag/1.14.2

PR #205, opened 2022-03-26
Remove Zend UriFactory

acairns:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | -
| Documentation   | -
| License         | MIT

#### What's in this PR?

PR is primarily context to explain proposal to remove Zend UriFactory support.

#### Why?

The Zend UriFactory is not PSR-17 compliant in the way that it generates it's URIs.

Example:
```
$uri = new \Zend\Diactoros\Uri('testing');
$this->assertEquals('testing', (string) $uri);
```
This test case would fail as the resulting Uri will actually be `/testing`.

Prepending a `/` is an assumption which has consequences - mentioned on`\Psr\Http\Message\UriInterface`

```
* Normally, the empty path "" and absolute path "/" are considered equal as
* defined in RFC 7230 Section 2.7.3. But this method MUST NOT automatically
* do this normalization because in contexts with a trimmed base path, e.g.
* the front controller, this difference becomes significant. It's the task
* of the user to handle both "" and "/".
```

More info here:
https://github.com/php-fig/http-message/blob/master/src/UriInterface.php#L134

As this UriFactory is prepending a `/` and not allowing the user to handle both `""` and `"/"`, different UriFactories can not be used interchangeably - preventing Liskov Substitution principle here.

Recently this caused us an issue where in many of our services, our new client operated as expected - primarily using versions of Guzzle. However, the Zend UriFactory was used in one of our legacy systems. Since our services have a path as part of their base_uri, this was stripped and our URIs were malformed due to prepending `/` - we addressed this issue by using [league/uri](https://github.com/thephpleague/uri).

#### Checklist

#### To Do

- [ ] The checklist above - primarily raising the PR to first discuss the approach and gain feedback

### DIFF
--- a/src/Strategy/CommonPsr17ClassesStrategy.php
+++ b/src/Strategy/CommonPsr17ClassesStrategy.php
@@ -78,7 +78,6 @@ final class CommonPsr17ClassesStrategy implements DiscoveryStrategy
         UriFactoryInterface::class => [
             'Phalcon\Http\Message\UriFactory',
             'Nyholm\Psr7\Factory\Psr17Factory',
-            'Zend\Diactoros\UriFactory',
             'GuzzleHttp\Psr7\HttpFactory',
             'Http\Factory\Diactoros\UriFactory',
             'Http\Factory\Guzzle\UriFactory',


dbu:
i am worried that this would lead to problems for existing users of discovery that had been using zend UriFactory and would end up with a system that does not discover a uri factory anymore after upgrading. 

zend has been renamed laminas (which we also support, but at lower priority). i guess zend things will not be updated anymore (and the change in the factory would be tricky as their users will have learned the wrong behaviour and rely on it...)? is laminas UriFactory correct?

we could do a new major version of discovery that drops no longer maintained providers like zend.

@Nyholm wdyt?

boesing:
@dbu you could add `zendframework/zend-diactoros` as `conflict` when you drop the interface. 

Laminas is out there for years now, even laminas dropped backwards compatibility layer for `zendframework` components while conflicting with them in several components while it `replace` the appropriate zendframework component in earlier versions:
https://github.com/laminas/laminas-diactoros/blob/b2355dc58253c480aa25ba085845a5aabf84955d/composer.json#L58 
https://github.com/laminas/laminas-diactoros/commit/8c644be420dd168f622d75af8cfd7eed5896c6f7

nicolas-grekas:
Closing in favor of #225, thanks for pushing this.

PR #207, opened 2022-07-11
Extend Exception from Throwable

gsteel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | None
| Documentation   | None required
| License         | MIT

#### What's in this PR?

Extends the Exception interface from Throwable

#### Why?

This stops static analysis from moaning about an invalid catch

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## 1.14.3 - Unreleased
+
+- [#207](https://github.com/php-http/discovery/pull/207) - Updates Exception to extend Throwable solving static analysis errors for consumers
+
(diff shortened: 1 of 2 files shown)


dbu (approved):
thanks for the PR. i don't think we have to worry about this "BC break", the interface is not supposed to be implemented from outside of this library.

what's more, if it would be possible to have an exception that does not implement `Throwable`, catching `Throwable` would not work with those exceptions, which would be problematic.

dbu:
https://github.com/php-http/discovery/releases/tag/1.14.3

PR #208, opened 2023-01-18
Add composer plugin to auto-install missing implementations

nicolas-grekas:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Documentation   | https://github.com/php-http/documentation/pull/301
| License         | MIT

This PR makes discovery able to auto-install a missing implementation when a dependency requires *both* php-http/discovery and one of the supported `*-implementation` virtual packages.

The code comes from https://github.com/FriendsOfPHP/well-known-implementations, which I plan to retire if this approach is accepted.

Note that we started discussing this topic with @dbu at SymfonyCon and we agreed this could be the best path forward.

#### To Do

- [x] Updated CHANGELOG.md to describe new feature
- [x] Documentation pull request created
- [x] Make the plugin work with Composer v1 and PHP 7.1
- [x] Add rule to install `http-interop/http-factory-guzzle` when Guzzle's PSR-7 **v1** is needed/installed and PSR-17 implementation is required
- [x] Add more test

### DIFF
--- a/.github/workflows/plugin.yml
+++ b/.github/workflows/plugin.yml
@@ -1,0 +1,36 @@
+name: Plugin
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  plugin:
+    name: Plugin test with Composer ${{ matrix.composer }}
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        include:
+          - composer: v1
+          - composer: v2
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: 7.1
+          tools: composer:${{ matrix.composer }}
+
+      - name: Check Plugin
+        run: |
+          mkdir /tmp/plugin
+          # replace the relative path for the repository url with an absolute path for composer v1 compatibility
+          jq '.repositories[0].url="'$(pwd)'"' tests/plugin/composer.json > /tmp/plugin/composer.json
+          cd /tmp/plugin
+          composer update
+          composer show http-interop/http-factory-guzzle -q
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,11 @@
 # Change Log
 
-## 1.15.0 - 2023-01-XX
+## 1.15.0 - 2023-02-09
 
 - [#209](https://github.com/php-http/discovery/pull/209) - Add generic `Psr17Factory` class
+- [#208](https://github.com/php-http/discovery/pull/208) - Add composer plugin to auto-install missing implementations.
+  When libraries require an http implementation but no packages providing that implementation is installed in the application, the plugin will automatically install one.
+  This is only done for libraries that directly require php-http/discovery to avoid unexpected dependency installation.
 
 ## 1.14.3 - 2022-07-11
 
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,7 @@
 {
     "name": "php-http/discovery",
-    "description": "Finds installed HTTPlug implementations and PSR-7 message factories",
+    "description": "Finds and installs PSR-7, PSR-17, PSR-18 and HTTPlug implementations",
+    "type": "composer-plugin",
     "license": "MIT",
     "keywords": ["http", "discovery", "client", "adapter", "message", "factory", "psr7", "psr17"],
     "homepage": "http://php-http.org",
@@ -10,18 +11,24 @@
             "email": "mark.sagikazar@gmail.com"
         }
     ],
+    "provide": {
+        "php-http/async-client-implementation": "*",
+        "php-http/client-implementation": "*",
+        "psr/http-client-implementation": "*",
+        "psr/http-factory-implementation": "*",
+        "psr/http-message-implementation": "*"
+    },
     "require": {
-        "php": "^7.1 || ^8.0"
+        "php": "^7.1 || ^8.0",
+        "composer-plugin-api": "^1.0|^2.0"
     },
     "require-dev": {
+        "composer/composer": "^1.0.2|^2.0",
         "graham-campbell/phpspec-skip-example-extension": "^5.0",
         "php-http/httplug": "^1.0 || ^2.0",
         "php-http/message-factory": "^1.0",
         "phpspec/phpspec": "^5.1 || ^6.1 || ^7.3",
         "symfony/phpunit-bridge": "^6.2"
-    },
-    "suggest": {
-        "php-http/message": "Allow to use Guzzle, Diactoros or Slim Framework factories"
     },
     "autoload": {
         "psr-4": {
@@ -39,6 +46,9 @@
             "vendor/bin/simple-phpunit --group NothingInstalled"
         ],
         "test-ci": "vendor/bin/phpspec run -c phpspec.ci.yml"
+    },
+    "extra": {
+        "class": "Http\\Discovery\\Composer\\Plugin"
     },
     "conflict": {
         "nyholm/psr7": "<1.0"
--- a/src/Composer/Plugin.php
+++ b/src/Composer/Plugin.php
@@ -1,0 +1,359 @@
+<?php
+
+namespace Http\Discovery\Composer;
+
+use Composer\Composer;
+use Composer\DependencyResolver\Pool;
+use Composer\EventDispatcher\EventSubscriberInterface;
+use Composer\Factory;
+use Composer\Installer;
+use Composer\IO\IOInterface;
+use Composer\Json\JsonFile;
+use Composer\Json\JsonManipulator;
+use Composer\Package\Locker;
+use Composer\Package\Version\VersionParser;
+use Composer\Package\Version\VersionSelector;
+use Composer\Plugin\PluginInterface;
+use Composer\Repository\InstalledRepositoryInterface;
+use Composer\Repository\RepositorySet;
+use Composer\Script\Event;
+use Composer\Script\ScriptEvents;
+
+/**
+ * Auto-installs missing implementations.
+ *
+ * When a dependency requires both this package and one of the supported `*-implementation`
+ * virtual packages, this plugin will auto-install a well-known implementation if none is
+ * found. The plugin will first look at already installed packages and figure out the
+ * preferred implementation to install based on the below stickyness rules (or on the first
+ * listed implementation if no rules match.)
+ *
+ * Don't miss updating src/Strategy/Common*Strategy.php when adding a new supported package.
+ *
+ * @author Nicolas Grekas <p@tchwork.com>
+ *
+ * @internal
+ */
+class Plugin implements PluginInterface, EventSubscriberInterface
+{
+    /**
+     * Describes, for every supported virtual implementation, which packages
+     * provide said implementation and which extra dependencies each package
+     * requires to provide the implementation.
+     */
+    private const PROVIDE_RULES = [
+        'php-http/async-client-implementation' => [
+            'symfony/http-client' => ['guzzlehttp/promises', 'php-http/message-factory', 'psr/http-factory-implementation'],
+            'php-http/guzzle7-adapter' => [],
+            'php-http/guzzle6-adapter' => [],
+            'php-http/curl-client' => [],
+            'php-http/react-adapter' => [],
+        ],
+        'php-http/client-implementation' => [
+            'symfony/http-client' => ['php-http/message-factory', 'psr/http-factory-implementation'],
+            'php-http/guzzle7-adapter' => [],
+            'php-http/guzzle6-adapter' => [],
+            'php-http/cakephp-adapter' => [],
+            'php-http/curl-client' => [],
+            'php-http/react-adapter' => [],
+            'php-http/buzz-adapter' => [],
+            'php-http/artax-adapter' => [],
+            'kriswallsmith/buzz:^1' => [],
+        ],
+        'psr/http-client-implementation' => [
+            'symfony/http-client' => ['psr/http-factory-implementation'],
+            'guzzlehttp/guzzle' => [],
+            'kriswallsmith/buzz:^1' => [],
+        ],
+        'psr/http-message-implementation' => [
+            'psr/http-factory-implementation' => [],
+        ],
+        'psr/http-factory-implementation' => [
+            'nyholm/psr7' => [],
+            'guzzlehttp/psr7:>=2' => [],
+            'slim/psr7' => [],
+            'laminas/laminas-diactoros' => [],
+            'phalcon/cphalcon:^4' => [],
+            'zendframework/zend-diactoros:>=2' => [],
+            'http-interop/http-factory-guzzle' => [],
+            'http-interop/http-factory-diactoros' => [],
+            'http-interop/http-factory-slim' => [],
+        ],
+    ];
+
+    /**
+     * Describes which package should be preferred on the left side
+     * depending on which one is already installed on the right side.
+     */
+    private const STICKYNESS_RULES = [
+        'symfony/http-client' => 'symfony/framework-bundle',
+        'php-http/guzzle7-adapter' => 'guzzlehttp/guzzle:^7',
+        'php-http/guzzle6-adapter' => 'guzzlehttp/guzzle:^6',
+        'php-http/guzzle5-adapter' => 'guzzlehttp/guzzle:^5',
+        'php-http/cakephp-adapter' => 'cakephp/cakephp',
+        'php-http/react-adapter' => 'react/event-loop',
+        'php-http/buzz-adapter' => 'kriswallsmith/buzz:^0.15.1',
+        'php-http/artax-adapter' => 'amphp/artax:^3',
+        'http-interop/http-factory-guzzle' => 'guzzlehttp/psr7:^1',
+        'http-interop/http-factory-diactoros' => 'zendframework/zend-diactoros:^1',
+        'http-interop/http-factory-slim' => 'slim/slim:^3',
+    ];
+
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            ScriptEvents::POST_UPDATE_CMD => 'postUpdate',
+        ];
+    }
+
+    public function activate(Composer $composer, IOInterface $io): void
+    {
+    }
+
+    public function deactivate(Composer $composer, IOInterface $io)
+    {
+    }
+
+    public function uninstall(Composer $composer, IOInterface $io)
+    {
+    }
+
+    public function postUpdate(Event $event)
+    {
+        $composer = $event->getComposer();
+        $repo = $composer->getRepositoryManager()->getLocalRepository();
+        $requires = [
+            $composer->getPackage()->getRequires(),
+            $composer->getPackage()->getDevRequires(),
+        ];
+
+        $missingRequires = $this->getMissingRequires($repo, $requires, 'project' === $composer->getPackage()->getType());
+        $missingRequires = [
+            'require' => array_fill_keys(array_merge([], ...array_values($missingRequires[0])), '*'),
+            'require-dev' => array_fill_keys(array_merge([], ...array_values($missingRequires[1])), '*'),
+            'remove' => array_fill_keys(array_merge([], ...array_values($missingRequires[2])), '*'),
+        ];
+
+        if (!$missingRequires = array_filter($missingRequires)) {
+            return;
+        }
+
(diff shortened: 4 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -26,13 +26,16 @@ jobs:
           coverage: none
 
       - name: Install PHP 7 dependencies
-        run: composer update --prefer-dist --no-interaction --no-progress
+        run: |
+          composer update --prefer-dist --no-interaction --no-progress
+          vendor/bin/simple-phpunit install
```
nicolas-grekas: The repo contains a few unit tests but they're never run by the CI, isn't it?
I'm adding simple-phpunit to run them - at least run the ones that don't require one of the well-know deps. This allows me to test the composer plugin using phpunit (I don't know phpspec, dunno if it's possible to use it for that. I'd need help if phpspec is a requirement.)

nicolas-grekas:
PR is ready. It comes with updates to the CI. (The remaining failures are unrelated)

.github/workflows/ci.yml
```diff
@@ -25,15 +25,10 @@ jobs:
...
-        run: composer update --prefer-dist --no-interaction --no-progress
-        if: "startsWith(matrix.php, '7.')"
-
-      - name: Install PHP 8 dependencies
+      - name: Install dependencies
         run: |
-          composer require "phpdocumentor/reflection-docblock:^5.2@dev" --no-interaction --no-update
-          composer update --prefer-dist --prefer-stable --no-interaction --no-progress --ignore-platform-req=php
-        if: "startsWith(matrix.php, '8.')"
+          composer update --prefer-dist --no-interaction --no-progress
```
dbu: thank you for this cleanup. seems this was for very bleeding edge php 8.0 testing.

.github/workflows/plugin.yml
```diff
@@ -0,0 +1,35 @@
...
+          php-version: 7.1
+          tools: composer:${{ matrix.composer }}
+
+      - name: Check Plugin
+        run: |
+          mkdir /tmp/plugin
+          jq '.repositories[0].url="'$(pwd)'"' tests/plugin/composer.json > /tmp/plugin/composer.json
+          cd /tmp/plugin
+          composer update
+          composer show http-interop/http-factory-guzzle -q
```
dbu, line 36: this will return a non 0 state if the guzzle factory is not installed?

.github/workflows/plugin.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: 7.1
+          tools: composer:${{ matrix.composer }}
+
+      - name: Check Plugin
+        run: |
+          mkdir /tmp/plugin
+          jq '.repositories[0].url="'$(pwd)'"' tests/plugin/composer.json > /tmp/plugin/composer.json
```
dbu, line 33: is this adding a repositories entry to composer.json? and if yes, why does that change make the plugin install the http factory?

composer.json
```diff
@@ -33,12 +41,16 @@
...
+            "vendor/bin/phpspec run",
+            "vendor/bin/simple-phpunit --group NothingInstalled"
+        ],
         "test-ci": "vendor/bin/phpspec run -c phpspec.ci.yml"
     },
     "extra": {
+        "class": "Http\\Discovery\\Composer\\Plugin",
         "branch-alias": {
-            "dev-master": "1.9-dev"
+            "dev-master": "1.15-dev"
```
dbu: i think we should change the branch to be 1.x and get rid of branch alias.

i will extract your general cleanup into a separate MR and merge that already to keep this separate.

src/Composer/Plugin.php
```diff
@@ -0,0 +1,347 @@
...
+ * @author Nicolas Grekas <p@tchwork.com>
+ *
+ * @internal
+ */
+class Plugin implements PluginInterface, EventSubscriberInterface
+{
+    /**
+     * Describes, for every supported virtual implementation, which packages
+     * provide said implementation and which extra dependencies each package
+     * requires to provide the implementation.
```
dbu, line 42: can you add to the documentation how the order of the providers matters?

the algorithm is something like checking if any of the implementations already exist (or is that handled by composer itself?) and then it tries the options in order to see if composer is happy with that option, and if not tries the next?

src/Composer/Plugin.php
```diff
@@ -0,0 +1,347 @@
...
+        'psr/http-factory-implementation' => [
+            'nyholm/psr7' => [],
+            'guzzlehttp/psr7:>=2' => [],
+            'slim/psr7' => [],
+            'laminas/laminas-diactoros' => [],
+            'phalcon/cphalcon:^4' => [],
+            'zendframework/zend-diactoros:>=2' => [],
+            'http-interop/http-factory-guzzle' => [],
+            'http-interop/http-factory-diactoros' => [],
+            'http-interop/http-factory-slim' => [],
```
dbu, line 80: these are similar to the lists in the strategy classes. i wonder if it should somehow be shared to avoid diverging data, like strategies finding implementations we don't try to autoinstall or worse, autoinstaller installing a valid implementation that discovery then does not find...

src/Composer/Plugin.php
```diff
@@ -0,0 +1,347 @@
...
+            return;
+        }
+
+        $composerJsonContents = file_get_contents(Factory::getComposerFile());
+        $this->updateComposerJson($missingRequires, $composer->getConfig()->get('sort-packages'));
+
+        $installer = null;
+        foreach (debug_backtrace(\DEBUG_BACKTRACE_PROVIDE_OBJECT) as $trace) {
+            if (isset($trace['object']) && $trace['object'] instanceof Installer) {
+                $installer = $trace['object'];
```
dbu, line 148: is there no more elegant way to access the installer of composer?

src/Composer/Plugin.php
```diff
@@ -0,0 +1,347 @@
...
+        $dispatcher = $composer->getEventDispatcher();
+        $disableScripts = !method_exists($dispatcher, 'setRunScripts') || !((array) $dispatcher)["\0*\0runScripts"];
+        $composer = Factory::create($event->getIO(), null, false, $disableScripts);
+
+        /** @var Installer $installer */
+        $installer = clone $installer;
+        if (method_exists($installer, 'setAudit')) {
+            $trace['object']->setAudit(false);
+        }
+        $installer->__construct(
```
dbu, line 169: why do we have to search the installer when we then overwrite its internals? (also, i was not aware you can call the constructor of an instantiated object, could we instead create a new instance, or is there more internal state on the installer that we need?)

src/Composer/Plugin.php
```diff
@@ -0,0 +1,347 @@
...
+            }
+        }
+
+        if ($updateComposerJson) {
+            $this->updateComposerJson($missingRequires, $composer->getConfig()->get('sort-packages'));
+            $this->updateComposerLock($composer, $event->getIO());
+        }
+    }
+
+    public function getMissingRequires(InstalledRepositoryInterface $repo, array $requires, bool $isProject): array
```
dbu, line 206: is this public for unit testing only? if so, we could also make it private and use reflection in the unit test to make it public

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change Log
 
+## 1.15.0 - 2023-01-XX
+
+- [#208](https://github.com/php-http/discovery/pull/208) - Add composer plugin to auto-install missing implementations
```
dbu:
```suggestion
- [#208](https://github.com/php-http/discovery/pull/208) - Add composer plugin to auto-install missing implementations if php-http/discovery is directly required in the application
```

want to defuse potential backlash against this functionality.

dbu (commented):
thanks a lot for the contribution! 

and thanks for the cleanup of the build system. i will extract those changes into a separate MR to fix the issues separately.

i have a few questions, mostly requesting more documentation to make the configuration of the plugin understandable without understanding the algorithm in its very details.

.github/workflows/ci.yml
dbu: i fear you are correct that we missed running the phpunit tests. phpspec is no requirement - the original authors started with phpspec, but we do have some phpunit tests and i myself am more familiar with that tool and don't know phpspec well.

dbu:
i merged the build setup cleanups to 1.x branch. please rebase this MR :-)

.github/workflows/plugin.yml
nicolas-grekas: absolutely!

src/Composer/Plugin.php
nicolas-grekas: nope, that's the only way... (this is borrowed from flex BTW)

src/Composer/Plugin.php
nicolas-grekas: the plugin is `@internal` anyway so I would prefer not if this argument works for you

.github/workflows/plugin.yml
nicolas-grekas: `tests/plugin/composer.json` looks like that:

```json

    "repositories": [
        {
            "type": "path",
            "url": "../..",
            "options": {
                "versions": {
                    "php-http/discovery": "99.99.x-dev"
                }
            }
        }
    ],
    "require": {
        "guzzlehttp/psr7": "^1",
        "php-http/discovery": "99.99.x-dev",
        "psr/http-factory-implementation": "*"
    },
    "config": {
        "allow-plugins": {
            "php-http/discovery": true
        }
    }
}
```

The fact that guzzle's psr7 *v1* is already required makes `http-interop/http-factory-guzzle` the best option to provide an implementation of `psr/http-factory-implementation`.

The change made with `jq` is to accommodate composer v1, who doesn't support having the repo url be a parent directory.

src/Composer/Plugin.php
nicolas-grekas: I prefer keeping these decoupled, because both descriptions don't have to be exactly the same. New vendors come up very rarely so I would suggest to deal with this syncing concern at the review stage.

CHANGELOG.md
nicolas-grekas: that wouldn't be accurate :)
the point of the plugin is precisely to allow libraries to *not* require any specific implementation and still be working out of the box, even if the root app doesn't have a matching implementation.
The requirement is that *libraries* must opt-in the plugin (apps can opt-out by deciding to disable the plugin or by requiring the implementation they prefer of course.)

nicolas-grekas: another way to opt-out for apps is just to require the implementation they prefer of course.

src/Composer/Plugin.php
nicolas-grekas: We do it this way because the constructor is not the only thing to call to configure the installer: the class has setters.
We want all the state of the existing installer (after all setters have been called) but we want a different set of service objects - the ones that the constructor needs.

src/Composer/Plugin.php
nicolas-grekas: I added this to the docblock of the class

nicolas-grekas:
PR rebased and green, comments addressed. Thanks for the review!

.github/workflows/plugin.yml
dbu: ah, sorry now i see. its not the composer.json of the library but a test fixture.

```suggestion
          # replace the relative path for the repository url with an absolute path for composer 1 compatibility.
          jq '.repositories[0].url="'$(pwd)'"' tests/plugin/composer.json > /tmp/plugin/composer.json
```

CHANGELOG.md
dbu: ah right.  would this be accurate?

```suggestion
- [#208](https://github.com/php-http/discovery/pull/208) - Add composer plugin to auto-install missing implementations.
  When libraries require a http implementation but no library providing that implementation is installed in the application, the plugin will automatically install an implementation.
  This is only done for libraries that directly require php-http/discovery to avoid unexpected dependency installation.
```

composer.json
```diff
@@ -10,19 +11,25 @@
             "email": "mark.sagikazar@gmail.com"
         }
     ],
+    "provide": {
+        "php-http/async-client-implementation": "*",
+        "php-http/client-implementation": "*",
+        "psr/http-client-implementation": "*",
+        "psr/http-factory-implementation": "*",
+        "psr/http-message-implementation": "*"
```
dbu, line 19: if the plugin is disabled or discovery only required indirectly, this would be incorrect, right? leading to confusing errors at runtime when discovery is used.

but i guess this can't be handled by the plugin, so we have to do it like this, right?

src/Composer/Plugin.php
```diff
@@ -0,0 +1,355 @@
...
+use Composer\Repository\RepositorySet;
+use Composer\Script\Event;
+use Composer\Script\ScriptEvents;
+
+/**
+ * Auto-installs missing implementations.
+ *
+ * When a dependency requires both this package and one of the supported `*-implementation`
+ * virtual packages, this plugin will auto-install a well-known implementation if none is
+ * found among the listed ones. The plugin will first look at already installed packages
```
dbu:
```suggestion
 * found in the list of well-known implementations. The plugin will first look at already installed packages
```

src/Composer/Plugin.php
```diff
@@ -0,0 +1,355 @@
...
+ * and figure out the preferred implementation to install based on the below stickyness
+ * rules (or on the first listed implementation if no rules apply.)
```
dbu:
```suggestion
 * rules (or on the first listed implementation if no rules match.)
```

src/Composer/Plugin.php
dbu: thanks!

src/Composer/Plugin.php
```diff
@@ -0,0 +1,355 @@
...
+
+        if (!$missingRequires = array_filter($missingRequires)) {
+            return;
+        }
+
+        $composerJsonContents = file_get_contents(Factory::getComposerFile());
+        $this->updateComposerJson($missingRequires, $composer->getConfig()->get('sort-packages'));
+
+        $installer = null;
+        foreach (debug_backtrace(\DEBUG_BACKTRACE_PROVIDE_OBJECT) as $trace) {
```
dbu, line 146:
```suggestion
        // Find the composer installer, hack borrowed from symfony/flex
        foreach (debug_backtrace(\DEBUG_BACKTRACE_PROVIDE_OBJECT) as $trace) {
```

src/Composer/Plugin.php
```diff
@@ -0,0 +1,355 @@
...
+        $dispatcher = $composer->getEventDispatcher();
+        $disableScripts = !method_exists($dispatcher, 'setRunScripts') || !((array) $dispatcher)["\0*\0runScripts"];
+        $composer = Factory::create($event->getIO(), null, false, $disableScripts);
+
+        /** @var Installer $installer */
+        $installer = clone $installer;
+        if (method_exists($installer, 'setAudit')) {
+            $trace['object']->setAudit(false);
+        }
+        $installer->__construct(
```
dbu, line 169:
```suggestion
        // we need a clone of the installer to have all its services that have been set, but with our own service objects
        $installer->__construct(
```

composer.json
nicolas-grekas: Correct if the plugin is disabled, but then ppl decided to do that. Existing apps won't be affected since they already have an implementation.
For indirect dep, the plugin will install the missing implem if any, so the declaration is still right.

nicolas-grekas:
Comments addressed

src/Composer/Plugin.php
dbu: could we cross-reference with comments to point to the other side? or use shared constants for the psr package names to highlight the relationship?

nicolas-grekas: sure, added

nicolas-grekas:
Rebased, all green, ready for merge + tag :)

dbu (approved):
thanks for the contribution

clxmstaab:
just to ping the relevant people: this PR lead to a breaking change for the current minor version which was reported in https://github.com/php-http/discovery/issues/213

no offence.

Tofandel:
It also has a bug, running composer install now with the plugin enabled leads to an infinite loop of composer install

src/Composer/Plugin.php
```diff
@@ -0,0 +1,359 @@
...
+        $missingRequires = $this->getMissingRequires($repo, $requires, 'project' === $composer->getPackage()->getType());
+        $missingRequires = [
+            'require' => array_fill_keys(array_merge([], ...array_values($missingRequires[0])), '*'),
+            'require-dev' => array_fill_keys(array_merge([], ...array_values($missingRequires[1])), '*'),
+            'remove' => array_fill_keys(array_merge([], ...array_values($missingRequires[2])), '*'),
+        ];
+
+        if (!$missingRequires = array_filter($missingRequires)) {
+            return;
+        }
```
Tofandel, line 139: Checking why it was looping, I did a var_dump of the `$missingRequires` and the keys of my `$requires`

```
array(22) {
  [0]=>
  string(3) "php"
  [1]=>
  string(12) "ext-fileinfo"
  [2]=>
  string(6) "ext-gd"
  [3]=>
  string(12) "ext-mbstring"
  [4]=>
  string(20) "based/momentum-modal"
  [5]=>
  string(29) "diglactic/laravel-breadcrumbs"
  [6]=>
  string(26) "donatello-za/rake-php-plus"
  [7]=>
  string(17) "guzzlehttp/guzzle"
  [8]=>
  string(25) "inertiajs/inertia-laravel"
  [9]=>
  string(17) "laravel/framework"
  [10]=>
  string(17) "laravel/jetstream"
  [11]=>
  string(15) "laravel/sanctum"
  [12]=>
  string(17) "maatwebsite/excel"
  [13]=>
  string(31) "psr/http-factory-implementation"
  [14]=>
  string(21) "sentry/sentry-laravel"
}
array(1) {
  ["require"]=>
  array(1) {
    ["psr/http-factory-implementation"]=>
    string(1) "*"
  }
}
```

So there is a logic flaw in the `getMissingRequires` as it's not removing the already existing requires of my composer.json

It seems `var_dump(array_keys($allPackages));` doesn't contain `psr/http-factory-implementation`

nicolas-grekas: @Tofandel can you let me know how to reproduce locally please?

src/Composer/Plugin.php
```diff
@@ -0,0 +1,359 @@
...
+        foreach (debug_backtrace(\DEBUG_BACKTRACE_PROVIDE_OBJECT) as $trace) {
+            if (isset($trace['object']) && $trace['object'] instanceof Installer) {
+                $installer = $trace['object'];
+                break;
+            }
+        }
+
+        if (!$installer) {
+            return;
+        }
```
Tofandel, line 155:
```php
        $isNestedInstall = true;
        $installer = null;
        // Find the composer installer, hack borrowed from symfony/flex
        foreach (debug_backtrace(\DEBUG_BACKTRACE_PROVIDE_OBJECT) as $trace) {
            if (isset($trace['object'])) {
                if ($trace['object'] instanceof Plugin) {
                    $isNestedInstall = false;
                }
                if ($trace['object'] instanceof Installer) {
                    $installer = $trace['object'];
                    break;
                }
            }
            if (isset($trace['object']) && $trace['object'] instanceof Installer) {
                $installer = $trace['object'];
                break;
            }
        }

        if (!$installer || $isNestedInstall) {
            return;
        }
```

Tofandel: To prevent the plugin from looping more than 2 times

src/Composer/Plugin.php
Tofandel: Sure just try to `composer upgrade` this `composer.json`

```json
{
  "name": "tofandel/example",
  "type": "project",
  "description": "",
  "require": {
    "php": "^8.1",
    "guzzlehttp/guzzle": "^7.0.1",
    "sentry/sentry-laravel": "*",
    "psr/http-factory-implementation": "*"
  },
  "require-dev": {
  },
  "config": {
    "allow-plugins": {
      "php-http/discovery": true
    }
  }
}
```

nicolas-grekas:
@Tofandel thanks for the reproducer, I opened https://github.com/php-http/discovery/pull/214 to fix the issue.

dunglas:
There is another issue with this plugin. If you pass options to `composer update` such as `--ignore-platform-reqs` they are discarded when the second update is discarded, and so the second update fails.

nicolas-grekas:
Good catch @dunglas. This should be fixed by https://github.com/php-http/discovery/pull/216

PR #209, opened 2023-01-27
Add generic Psr17Factory

nicolas-grekas:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| Documentation   | https://github.com/php-http/documentation/pull/303
| License         | MIT

This PR ports [WellKnownPsr17Factory](https://github.com/FriendsOfPHP/well-known-implementations/blob/main/src/WellKnownPsr17Factory.php) to this repository.

This approach should make it even simpler to write generic PSR17-consuming code.
It would also make the package compatible with factories declared as default values.

Last but not least, the implementation in Psr17Factory provides two additional methods
- `Psr17Factory::createServerRequestFromGlobals()`
- `Psr17Factory::createUriFromGlobals()`

Providing these two methods provides true portability of PSR-7 implementations.
An example that shows this is missing is Sentry's [`RequestFetcher`](https://github.com/getsentry/sentry-php/blob/master/src/Integration/RequestFetcher.php), which is currently coupled to `guzzlehttp/psr7` because of the call to `ServerRequest::fromGlobals()`.

#### Example Usage

Inline:
``` php
$factory = new Psr17Factory();

// Now we can call any PSR-17 method, e.g.
$factory->createRequest();
```

As default value on PHP8+:
```php
class Foo
{
    public function __construct(
        private RequestFactoryInterface $requestFactory = new Psr17Factory(),
    ) {
        //...
    }
}
```

#### To Do

- [x] Updated CHANGELOG.md
- [x] Documentation pull request created
- [x] Fix tests

### DIFF
--- a/src/Psr17Factory.php
+++ b/src/Psr17Factory.php
@@ -1,0 +1,282 @@
+<?php
+
+namespace Http\Discovery;
+
+use Psr\Http\Message\RequestFactoryInterface;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseFactoryInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestFactoryInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Message\StreamFactoryInterface;
+use Psr\Http\Message\StreamInterface;
+use Psr\Http\Message\UploadedFileFactoryInterface;
+use Psr\Http\Message\UploadedFileInterface;
+use Psr\Http\Message\UriFactoryInterface;
+use Psr\Http\Message\UriInterface;
+
+/**
+ * A generic PSR-17 implementation.
+ *
+ * You can create this class with concrete factory instances or let
+ * it use discovery to find suitable implementations as needed.
+ *
+ * This class also provides two additional methods that are not in PSR17,
+ * to help with creating PSR-7 objects from PHP superglobals:
+ *  - createServerRequestFromGlobals()
+ *  - createUriFromGlobals()
+ *
+ * The code in this class is inspired by the "nyholm/psr7", "guzzlehttp/psr7"
+ * and "symfony/http-foundation" packages, all licenced under MIT.
+ *
+ * Copyright (c) 2004-2023 Fabien Potencier <fabien@symfony.com>
+ * Copyright (c) 2015 Michael Dowling <mtdowling@gmail.com>
+ * Copyright (c) 2015 Márk Sági-Kazár <mark.sagikazar@gmail.com>
+ * Copyright (c) 2015 Graham Campbell <hello@gjcampbell.co.uk>
+ * Copyright (c) 2016 Tobias Schultze <webmaster@tubo-world.de>
+ * Copyright (c) 2016 George Mponos <gmponos@gmail.com>
+ * Copyright (c) 2016-2018 Tobias Nyholm <tobias.nyholm@gmail.com>
+ *
+ * @author Nicolas Grekas <p@tchwork.com>
+ */
+class Psr17Factory implements RequestFactoryInterface, ResponseFactoryInterface, ServerRequestFactoryInterface, StreamFactoryInterface, UploadedFileFactoryInterface, UriFactoryInterface
+{
+    private $requestFactory;
+    private $responseFactory;
+    private $serverRequestFactory;
+    private $streamFactory;
+    private $uploadedFileFactory;
+    private $uriFactory;
+
+    public function __construct(
+        RequestFactoryInterface $requestFactory = null,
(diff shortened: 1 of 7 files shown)


src/Psr17Factory.php
```diff
@@ -0,0 +1,278 @@
...
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Message\StreamFactoryInterface;
+use Psr\Http\Message\StreamInterface;
+use Psr\Http\Message\UploadedFileFactoryInterface;
+use Psr\Http\Message\UploadedFileInterface;
+use Psr\Http\Message\UriFactoryInterface;
+use Psr\Http\Message\UriInterface;
+
+/**
+ * A portable PSR-17 implementation.
```
dbu: could we call this a convenience class that provides all the various psr17 factories in one class?

src/Psr17Factory.php
```diff
@@ -0,0 +1,278 @@
...
+     * @param UriInterface|string $uri
+     */
+    public function createServerRequest(string $method, $uri, array $serverParams = []): ServerRequestInterface
+    {
+        $factory = $this->serverRequestFactory ?? $this->setFactory(Psr17FactoryDiscovery::findServerRequestFactory());
+
+        return $factory->createServerRequest(...\func_get_args());
+    }
+
+    public function createServerRequestFromGlobals(array $server = null, array $get = null, array $post = null, array $cookie = null, array $files = null, StreamInterface $body = null): ServerRequestInterface
```
dbu, line 101: should we mention that this is not part of psr17 but a utility method?

src/Psr17Factory.php
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    public function createUri(string $uri = ''): UriInterface
+    {
+        $factory = $this->uriFactory ?? $this->setFactory(Psr17FactoryDiscovery::findUriFactory());
+
+        return $factory->createUri(...\func_get_args());
+    }
+
+    public function createUriFromGlobals(array $server = null): UriInterface
```
dbu, line 151: same question here, mention that this is a convenience function that is not part of psr17?

dbu (commented):
interesting. it does not directly have to do with discovery, not convinced if it fits into this repository.

looking at what we currently provide, i don't see a clear home for this. but like we have php-http/message about psr 7 things, we could do a php-http/factory repository for it?

i think i would prefer that over having it as part of discovery. wdyt? i could set up that repository and give you commit rights if you want.

src/Psr17Factory.php
nicolas-grekas: I added a note about this in the docblock of the class

src/Psr17Factory.php
nicolas-grekas: I added a note about this in the docblock of the class

src/Psr17Factory.php
nicolas-grekas: It a bit more than that: all known PRS-17 implementations provide all methods in one class, and this one is no different.
I changed the comment to "A generic PSR-17 implementation.", since generic is really what this is about: some independent from the actual implementation.

nicolas-grekas:
> it does not directly have to do with discovery, not convinced if it fits into this repository.

Arf. To me, discovery is the package where this makes most sense, because that class is using discovery, and because it fills the same need - decoupling to any specific PSR-17 implem. Making a separate package would just make maintenance harder and slow down adoption to me.

dbu:
thanks for the discussion in slack @nicolas-grekas , now i got what it does. its a decorator around the `findX` methods that provides you with an all-of-psr17 factory that autodiscovers its things if needed but accepts constructor arguments when autodiscovery is not wanted.

yep, that fits into this package then.

src/Psr17Factory.php
```diff
@@ -0,0 +1,283 @@
...
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Message\StreamFactoryInterface;
+use Psr\Http\Message\StreamInterface;
+use Psr\Http\Message\UploadedFileFactoryInterface;
+use Psr\Http\Message\UploadedFileInterface;
+use Psr\Http\Message\UriFactoryInterface;
+use Psr\Http\Message\UriInterface;
+
+/**
+ * A generic PSR-17 implementation.
```
dbu, line 19:
```suggestion
 * A generic PSR-17 implementation.
 *
 * You can create this class with concrete factory instances or let it use discovery to find suitable implementations as needed.
```

nicolas-grekas: added and pushed thanks

nicolas-grekas:
I just pushed tests for the `*FromGlobals()` methods. PR is ready on my side.

dbu (approved):
great, thanks!

PR #219, opened 2023-02-11
Fix handling of replaced packages

### LINKED ISSUE #213: 1.15.0 breaks CI pipelines
**PHP version**: 8.1.15
**Composer version**: 2.5.2
**Description**

Composer asks if I want to trust "php-http/discovery" when 1.15.0 gets required. This breaks CI pipelines when php-http/discovery is an indirect dependency.

**How to reproduce**
composer require php-http/discovery

**Possible Solution**
Use a new major version  (2.0.0) for breaking changes.

**Additional context**
$ composer require php-http/discovery
./composer.json has been created
Running composer update php-http/discovery
Loading composer repositories with package information
Updating dependencies
Lock file operations: 1 install, 0 updates, 0 removals
  - Locking php-http/discovery (1.15.0)
Writing lock file
Installing dependencies from lock file (including require-dev)
Package operations: 1 install, 0 updates, 0 removals
php-http/discovery contains a Composer plugin which is currently not in your allow-plugins config. See https://getcomposer.org/allow-plugins
Do you trust "php-http/discovery" to execute code and wish to enable it now? (writes "allow-plugins" to composer.json) [y,n,d,?]

nicolas-grekas:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes https://github.com/php-http/discovery/issues/213#issuecomment-1426132264
| Documentation   | -
| License         | MIT

Solves https://stackoverflow.com/questions/75414421/is-php-http-discovery-plugin-1-15-1-still-breaking-composer-install-in-contao-4 by taking care of "replace" rules in installed packages.

### DIFF
--- a/src/Composer/Plugin.php
+++ b/src/Composer/Plugin.php
@@ -209,18 +209,26 @@ class Plugin implements PluginInterface, EventSubscriberInterface
     public function getMissingRequires(InstalledRepositoryInterface $repo, array $requires, bool $isProject): array
     {
         $allPackages = [];
-        $devPackages = method_exists($repo, 'getDevPackageNames') ? array_flip($repo->getDevPackageNames()) : [];
+        $devPackages = method_exists($repo, 'getDevPackageNames') ? array_fill_keys($repo->getDevPackageNames(), true) : [];
 
-        // One must require "php-http/discovery" or "friendsofphp/well-known-implementations"
+        // One must require "php-http/discovery"
         // to opt-in for auto-installation of virtual package implementations
-        if (!isset($requires[0]['php-http/discovery']) && !isset($requires[0]['friendsofphp/well-known-implementations'])) {
+        if (!isset($requires[0]['php-http/discovery'])) {
             $requires = [[], []];
         }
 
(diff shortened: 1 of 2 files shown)


amenk:
Confirmed working. Thank you :-)

vahidalvandi:
also i have error

  php-http/discovery contains a Composer plugin which is blocked by your allow-plugins config. You may add it to the list if you consider it safe.         
  You can run "composer config --no-plugins allow-plugins.php-http/discovery [true|false]" to enable it (true) or disable it explicitly and suppress this  
   exception (false)                                                                                                                                       
  See https://getcomposer.org/allow-plugins

dbu:
@vahidalvandi this is no error, but an information. and the information explains you what you can do to not see the message anymore, as well as a link to the composer information on what plugins are.

the plugin would automatically install a suitable PSR-7/PSR-17/PSR-18 implementation if one is required but none explicitly selected by your application.

PR #232, opened 2023-04-26
Allow pinning the preferred implementations in composer.json

### LINKED ISSUE #223: Allow pinning an implementation when several are available
When `vendor/` contains several implementations of e.g. PSR-17, the discovery classes will select one based on some internal order of preference.

But the author of an app should be able to decide which implementation should be preferred.

There are already static methods to configure this, but it'd be great to allow specifying this in composer.json files.

To make this really fast, reading this value should be handled at "composer install" time, when the plugin can generate an ad hoc config file in PHP.

nicolas-grekas:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #223
| Documentation   | TBD
| License         | MIT

Add this to your composer.json to pin guzzle as PSR-17 implementation even if nyholm/psr7 is installed:
```json
    "extra": {
        "discovery": {
            "psr/http-factory-implementation": "GuzzleHttp\\Psr7\\HttpFactory"
        }
    }
```

This also works for single interfaces:
```json
    "extra": {
        "discovery": {
            "Psr\\Http\\Message\\RequestFactoryInterface": "Slim\\Psr7\\Factory\\RequestFactory"
        }
    }
```

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -13,7 +13,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Roave BC Check
         uses: "docker://nyholm/roave-bc-check-ga"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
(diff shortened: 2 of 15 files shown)


boesing:
Huge! Thanks.

dbu:
i am adding some documentation for this in https://github.com/php-http/documentation/pull/305

dbu:
d'oh, i missed https://github.com/php-http/documentation/pull/304/

PR #239, opened 2023-05-03
Skip auto-installing when the root package's extra.discovery is enough

### LINKED ISSUE #237: Skip auto-installing when the root package's extra.discovery is enough
Here is an idea that popped up in #236 and that builds on #232.

When the root package configures this:
```json
    "extra": {
        "discovery": {
            "psr/http-factory-implementation": "My\\Psr17Factory"
        }
    }
```

then we shouldn't need auto-installing any extra packages for this virtual package.

Same when listing individual interfaces:
```json
    "extra": {
        "discovery": {
            "Psr\\Http\\Client\\ClientInterface": "My\\Psr18Client"
        }
    }
```

In this case, we should verify that all the interfaces required to fulfill one of the supported `*-implementation` are listed.

nicolas-grekas:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #237
| Documentation   | -
| License         | MIT

### DIFF
--- a/.github/workflows/plugin.yml
+++ b/.github/workflows/plugin.yml
@@ -42,4 +42,4 @@ jobs:
           jq '.repositories[0].url="'$(pwd)'"' tests/plugin/pinning/composer.json > /tmp/plugin-pinning/composer.json
           cd /tmp/plugin-pinning
           composer update
-          [ 'Slim\Psr7\Factory\RequestFactory' == $(php test.php) ]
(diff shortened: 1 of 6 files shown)


dbu (commented):
neat!

should we expand https://github.com/php-http/documentation/blob/master/discovery.rst a bit to explain this in the autoinstallation section?

PR #243, opened 2023-05-17
Update README example to work in ZSH

donatj:
| Q               | A
| --------------- | ---
| Bug fix?        | yes?
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | no
| Documentation   | yes
| License         | MIT

#### What's in this PR?

Just quotes the example in the README

#### Why?

zsh and likely some other shells try to glob the * and cause the lines to fail.

image

zsh being the default shell on macOS now, seems like it should get fixed.

I quoted the first option for visual symmetry but it doesn't really require it.

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,9 @@ Finally, you need to require `php-http/discovery` and the generic implementation
 that your library is going to need:
 
 ```bash
-composer require php-http/discovery:^1.17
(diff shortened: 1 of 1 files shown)


dbu (approved):
thanks!

PR #249, opened 2023-06-17
add missing requirements of symfony/http-client to composer plugin

### LINKED ISSUE #248: You cannot use the "Symfony\Component\HttpClient\Psr18Client" as the "psr/http-client" package is not installed
**PHP version**: 8.2.0

**Description**
php-http/discovery plugin installs a psr18 client that cannot be discovered

**How to reproduce**
composer.json:
```json
{
  "config": {
    "allow-plugins": {
      "php-http/discovery": true
    }
  },
  "require": {
    "php-http/client-implementation": "^1.0",
    "php-http/discovery": "^1.18"
  }
}
```

test.php:
```php
<?php
require 'vendor/autoload.php';
$client = \Http\Discovery\Psr18ClientDiscovery::find();
var_dump(get_class($client));
```

```shell
composer update
php test.php
```

Results:
```shell
Fatal error: Uncaught LogicException: You cannot use the "Symfony\Component\HttpClient\Psr18Client" as the "psr/http-client" package is not installed. Try running "composer require php-http/discovery psr/http-client-implementation:*". in /usr/src/myapp/vendor/symfony/http-client/Psr18Client.php on line 37
```
**Possible Solution**
As a workaround, I could require `psr/http-client`, but we don't directly require or use it.

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #248
| Documentation   | -
| License         | MIT

#### What's in this PR?

As discovered in #248, the composer plugin does not install a functional set of dependencies to actually discover the psr18 client when installing symfony/http-client.

#### Reproduce

To reproduce the problem, create an empty folder, create `composer.json` with the following content and run `composer update`:

```
{
  "config": {
    "allow-plugins": {
      "php-http/discovery": true
    }
  },
  "require": {
    "psr/http-client-implementation": "*",
    "php-http/discovery": "^1.18"
  }
}
```

`ls vendor/psr` shows that psr/http-client is not installed.

#### Checklist

### DIFF
--- a/src/Composer/Plugin.php
+++ b/src/Composer/Plugin.php
@@ -45,16 +45,16 @@ class Plugin implements PluginInterface, EventSubscriberInterface
      */
     private const PROVIDE_RULES = [
         'php-http/async-client-implementation' => [
-            'symfony/http-client:>=6.3' => ['guzzlehttp/promises', 'psr/http-factory-implementation'],
-            'symfony/http-client' => ['guzzlehttp/promises', 'php-http/message-factory', 'psr/http-factory-implementation'],
+            'symfony/http-client:>=6.3' => ['guzzlehttp/promises', 'psr/http-factory-implementation', 'php-http/httplug'],
+            'symfony/http-client' => ['guzzlehttp/promises', 'php-http/message-factory', 'psr/http-factory-implementation', 'php-http/httplug'],
             'php-http/guzzle7-adapter' => [],
             'php-http/guzzle6-adapter' => [],
             'php-http/curl-client' => [],
             'php-http/react-adapter' => [],
         ],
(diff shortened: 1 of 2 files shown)


dbu:
@nicolas-grekas i will adjust the unit tests and update changelog if you agree that this is the correct thing to do.

nicolas-grekas (approved):
Looks correct to me. I suppose I was expecting that libs that require an implem would also require the abstraction :)

dbu:
> I suppose I was expecting that libs that require an implem would also require the abstraction :)

the trap of optional dependencies... i adjusted the test expectation.

PR #250, opened 2023-07-10
fix: buzz client instantiation using deprecated Message Factory Discovery

mnavarrocarter:
Fixes the instantiation of Buzz client to use the new `Psr17FactoryDiscovery` instead of the deprecated `MessageFactoryDiscovery`.

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no|
| Deprecations?   | no
| Related tickets | none
| Documentation   | none
| License         | MIT

#### What's in this PR?

Change Buzz instantiator from the deprecated `MessageFactoryDisvovery` to the new `Psr17FactoryDiscovery`. 

#### Why?

The Buzz Instantiator was using the deprectaed `MessageFactoryDiscovery` and this was causing an exception when attempting to instantiate Buzz using `nyholm/psr7`. Changed to the more modern `Psr17FactoryDiscovery` that does find `nyholm/psr7` for Buzz.

#### Example Usage
Noop

#### Checklist

### DIFF
--- a/src/Strategy/CommonClassesStrategy.php
+++ b/src/Strategy/CommonClassesStrategy.php
@@ -18,7 +18,6 @@ use Http\Client\HttpClient;
 use Http\Client\Socket\Client as Socket;
 use Http\Discovery\ClassDiscovery;
 use Http\Discovery\Exception\NotFoundException;
-use Http\Discovery\MessageFactoryDiscovery;
 use Http\Discovery\Psr17FactoryDiscovery;
 use Http\Message\MessageFactory;
 use Http\Message\MessageFactory\DiactorosMessageFactory;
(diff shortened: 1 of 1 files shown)


dbu (approved):
we seem to not have an [installation test](https://github.com/php-http/discovery/blob/1.x/.github/workflows/installation.yml) with buzz but checking the code, the [client accepts both factories since 1.0.0](https://github.com/kriswallsmith/Buzz/blob/1.0.0/lib/Client/AbstractClient.php)

dbu:
thanks!
https://github.com/php-http/discovery/releases/tag/1.19.1

PR #251, opened 2023-07-27
fix:replace deprecated `HttplugFactory` with `Psr17Factory`

BadJacky:
# Replace deprecated HttplugFactory with Psr17Factory

## Description

HttplugFactory has been deprecated since version 1.8. 

This PR replaces all usage of HttplugFactory with the recommended Psr17Factory class.

## Changes

- Replaced all instances of "HttplugFactory" with "Psr17Factory"
- Removed deprecated HttplugFactory class uses

## Checklist

- [ ] Updated CHANGELOG.md
- [ ] Updated documentation

## Additional Information

HttplugFactory has been deprecated since version 1.8. Using the deprecated class would cause deprecation warnings. This PR replaces all usage with the recommended Psr17Factory class to avoid warnings and use the latest implementation.

The deprecated HttplugFactory class has been completely removed after replacing all usages.

### DIFF
--- a/src/Strategy/CommonClassesStrategy.php
+++ b/src/Strategy/CommonClassesStrategy.php
@@ -32,7 +32,7 @@ use Http\Message\UriFactory\DiactorosUriFactory;
 use Http\Message\UriFactory\GuzzleUriFactory;
 use Http\Message\UriFactory\SlimUriFactory;
 use Laminas\Diactoros\Request as DiactorosRequest;
-use Nyholm\Psr7\Factory\HttplugFactory as NyholmHttplugFactory;
+use Nyholm\Psr7\Factory\Psr17Factory as NyholmPsr17Factory;
 use Psr\Http\Client\ClientInterface as Psr18Client;
 use Psr\Http\Message\RequestFactoryInterface as Psr17RequestFactory;
 use Slim\Http\Request as SlimRequest;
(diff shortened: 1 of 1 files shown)


src/Strategy/CommonClassesStrategy.php
```diff
@@ -53,19 +53,19 @@ final class CommonClassesStrategy implements DiscoveryStrategy
      */
     private static $classes = [
         MessageFactory::class => [
-            ['class' => NyholmHttplugFactory::class, 'condition' => [NyholmHttplugFactory::class]],
+            ['class' => NyholmPsr17Factory::class, 'condition' => [NyholmPsr17Factory::class]],
```
dbu, line 56: in the interest of BC with older nyholm libraries, should we keep the legacy factory as second option, after the psr17 factory?

dbu: ah okay, https://github.com/Nyholm/psr7/blob/1.0.0/src/Factory/Psr17Factory.php says the factory is available in all versions of nyholm psr7

dbu:
i think there is a misunderstanding. the Psr7Factory is for the PSR-17 factories, the message factories are for https://github.com/php-http/message factories. the 3 things you change are to discover the legacy php-http factories. the psr7 factory is not a factory for php-http/message things.

people should not try to discover those legacy php-http factories anymore, but the psr17 factories. those are defined in https://github.com/php-http/discovery/blob/1.x/src/Strategy/CommonPsr17ClassesStrategy.php

BadJacky:
> i think there is a misunderstanding. the Psr7Factory is for the PSR-17 factories, the message factories are for https://github.com/php-http/message factories. the 3 things you change are to discover the legacy php-http factories. the psr7 factory is not a factory for php-http/message things.
> 
> people should not try to discover those legacy php-http factories anymore, but the psr17 factories. those are defined in https://github.com/php-http/discovery/blob/1.x/src/Strategy/CommonPsr17ClassesStrategy.php

I believe I've grasped your intentions.

Upon inspecting the Nyholm/Psr7 package, I have concluded that this does not present a backward compatibility issue.

Additionally, I wanted to inquire about our current use of deprecated classes within the project. Could we possibly consider eliminating them?

dbu:
one should not use MessageFactoryDiscovery / MessageFactoryDiscovery / UriFactoryDiscovery anymore, but the Psr17FactoryDiscovery to get psr17 factories for those psr7 items. the old factories had been done when psr17 did not exist yet and have all been deprecated in favor of psr17

PR #253, opened 2023-11-30
Check if SymfonyHttplug actually implements HttpClient

### LINKED ISSUE #252: Symfony 7 no longer implements Http\Client\HttpClient
**Description**
symfony/http-client no longer implements Http\Client\HttpClient so discovery should not return it.

**How to reproduce**
run 
```php
<?php
var_dump(\Http\Discovery\HttpClientDiscovery::find() instanceof  Http\Client\HttpClient); // false
```
with symfony/http-client v7 installed

arokettu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #252
| License         | MIT

#### What's in this PR?

When looking for HttpClient, check if SymfonyHttplug actually implements HttpClient.

#### Why?

In Symfony 7, SymfonyHttplug no longer implements HttpClient

#### Checklist

### DIFF
--- a/src/Strategy/CommonClassesStrategy.php
+++ b/src/Strategy/CommonClassesStrategy.php
@@ -160,7 +160,7 @@ final class CommonClassesStrategy implements DiscoveryStrategy
 
     public static function isSymfonyImplementingHttpClient()
     {
-        return is_a(SymfonyHttplug::class, HttpClient::class, true);
+        return is_subclass_of(SymfonyHttplug::class, HttpClient::class);
     }
 
     /**
--- a/src/Strategy/CommonClassesStrategy.php
+++ b/src/Strategy/CommonClassesStrategy.php
@@ -78,7 +78,7 @@ final class CommonClassesStrategy implements DiscoveryStrategy
             ['class' => React::class, 'condition' => React::class],
         ],
         HttpClient::class => [
-            ['class' => SymfonyHttplug::class, 'condition' => [SymfonyHttplug::class, [self::class, 'isPsr17FactoryInstalled']]],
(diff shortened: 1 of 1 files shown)


src/Strategy/CommonClassesStrategy.php
```diff
@@ -158,6 +158,11 @@ public static function isGuzzleImplementingPsr18()
         return defined('GuzzleHttp\ClientInterface::MAJOR_VERSION');
     }
 
+    public static function isSymfonyImplementingHttpClient()
```
nicolas-grekas, line 161:
```suggestion
    public static function isSymfonyImplementingHttplug()
```

src/Strategy/CommonClassesStrategy.php
```diff
@@ -158,6 +158,11 @@ public static function isGuzzleImplementingPsr18()
...
+    {
+        return is_a(SymfonyHttplug::class, HttpClient::class, true);
```
nicolas-grekas:
```suggestion
        return is_subclass_of(SymfonyHttplug::class, HttpClient::class);
```

src/Strategy/CommonClassesStrategy.php
arokettu: HttpAsyncClient is kinda Httplug too

dbu (approved):
thanks a lot for this one.

src/Strategy/CommonClassesStrategy.php
dbu: i agree with @arokettu 

its not going to be relevant forever, we deprecated the synchronous php-http HttpClient interface in favor of the PSR18 http client. only the async php-http client is still relevant. when we do a new major of discovery, we will drop discovering HttpClient.

dbu:
https://github.com/php-http/discovery/releases/tag/1.19.2

PR #263, opened 2024-03-28
WIP: make sure we install php-interop 1 when asking for the psr factory

dbu:
the 0.x versions are for the php-http factories

### DIFF
--- a/src/Composer/Plugin.php
+++ b/src/Composer/Plugin.php
@@ -78,7 +78,7 @@ class Plugin implements PluginInterface, EventSubscriberInterface
             'slim/psr7' => [],
             'laminas/laminas-diactoros' => [],
             'phalcon/cphalcon:^4' => [],
-            'http-interop/http-factory-guzzle' => [],
+            'http-interop/http-factory-guzzle:>=1' => [],
             'http-interop/http-factory-diactoros' => [],
             'http-interop/http-factory-slim' => [],
             'httpsoft/http-message' => [],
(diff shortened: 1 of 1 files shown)


src/Composer/Plugin.php
```diff
@@ -78,7 +78,7 @@ class Plugin implements PluginInterface, EventSubscriberInterface
             'slim/psr7' => [],
             'laminas/laminas-diactoros' => [],
             'phalcon/cphalcon:^4' => [],
-            'http-interop/http-factory-guzzle' => [],
+            'http-interop/http-factory-guzzle:>=1' => [],
```
dbu, line 81: @nicolas-grekas is this not the correct way to do this? i noticed in the FOSHttpCache --prefer-lowest installs that i end up with http-interop/http-factory-guzzle 0.2.0 which is for the legacy php-http factories and not the psr factories.

when i use this branch, the plugin adds a require for `http-interop/http-factory-guzzle: *` and then goes into an endless loop. i accidentally created a reproducer in https://github.com/FriendsOfSymfony/FOSHttpCache/pull/567

nicolas-grekas: I had a look at the code and we don't support version constraints on the auto-installed packages.
Instead, conflict rules should be used on the consuming side (so that the correct dep is installed with or without the plugin)

dbu: the problem is that `http-interop/http-factory-guzzle` is only a `psr/http-factory-implementation` if we use at least version 1. we have version constraints on other packages in this list, just above, e.g. phalcon or guzzlehttp/psr7. would those also break if an older version was to be installed?

nicolas-grekas: The other version constraints apply on the "what's already installed" side, not on the "what should be installed" one...

dbu:
thanks for the clarification.

not perfect, but prefer lowest is a CI question rather than a real world use case. and with minimal php versions increasing, this particular issue will fade out. fixing this on the FOSHttpCache side for now.

PR #266, opened 2024-04-22
Forward factories implemented by PSR-18 clients if any

nicolas-grekas:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| Documentation   | -
| License         | MIT

E.g. Symfony's PSR-18 client implements the relevant PSR-17 factories.
This is just a minor improvement.


dbu (approved):
was wondering why we don't do that in Psr17Factory instead, but we don't pass the ClientInterface to it so thats why...

it makes sense to do that, yes.

